PR #1, opened 2016-12-15
Add zmq-publishers to nodepool

jamielennox:
Hey! What do you know, there's another communication mechanism between
nodepool and zuul where nodepool listens for things like onFinalized
which is the thing that deletes the nodes when a job is finished.

Seriously. We need to get this under control.


ConsoleCatzirl (approved):
LGTM

PR #2, opened 2016-12-15
Require DCO for all contributions

nibalizer:


missaugustina (changes requested):
This is awesome and explains everything really well. I have a few things I'd like to see changed:
1. We should make a CONTRIBUTING.rst document and move this content in there
2. It needs instructions on how to do this in git
3. We should talk about automagic enforcement of this as a team and merge whatever mechanism we decide to use for that at the same time as this change.

rattboi (changes requested):
I'd prefer if this were written in a way where it was assumed that success was false, and if it passes some test, then success is set to true. 

The likelihood of some corner case allowing through unsigned commits is a lot more narrow that way.

test/signed-off-by-test.sh
```diff
@@ -0,0 +1,23 @@
...
+        git show $commit
+        success="false"
+    fi
+
+done
+
+if [ "$success" = "true" ]; then
+    echo "All commits have a Signed-off-by"
+else
+    echo "Some commits do not have a signed off by, failing"
```
rattboi: It would be better as Signed-off-by for consistency.

Maybe even 'Signed-off-by (git commit -s)' instead, so that it's extremely obvious to new contributors.

rattboi (changes requested):

rattboi:
https://travis-ci.org/BonnyCI/hoist/jobs/184587798

Looks like maybe the sha has drifted because this hasn't merged, or there's some other issue we should look into.

CONTRIBUTING.md
```diff
@@ -0,0 +1,56 @@
...
+between the project and the contributor, we've introduced a "sign-off"
+procedure on patches/pull requests submitted to this project.
+
+The sign-off is a simple line at the end of the explanation for the
+patch, which certifies that you wrote it or otherwise have the right to
+pass it on as an open-source patch.  The rules are pretty simple: if you
+can certify the below:
+
+Developer's Certificate of Origin 1.1
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
omgjlk: This line of carrots does not render well on github's markdown.

eggshell:
@rattboi looking at the git log, `c10f159` doesn't have a `Signed-off-by`.

omgjlk (commented):
I only object to the rendering of carrots.

test/signed-off-by-test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+first_signed_off_commit="93a80b421560de540ecaf85850074e851cecf2b1"
+success="true"
+
+for commit in $(git log --pretty=oneline $first_signed_off_commit..HEAD | cut -d " " -f 1)
+do
+    if $(git show $commit | grep -q Signed-off-by ); then
```
eggshell, line 8: Might want to make sure this isn't checking merge commits.

missaugustina (approved):

missaugustina:
I am approved as soon as the thing works ;) Thanks for doing this!

PR #4, opened 2016-12-15
Set zuul merger url

jamielennox:
The zuul merger clones the base repository from the original source to
take load off the merger servers and then takes the change ref from the
merger. We need to point the zuul merger url at this url.


ConsoleCatzirl (approved):
For now this is good, but when we start splitting out mergers to scale them out we will need a way to set this to the hostname for each merger.

PR #6, opened 2016-12-15
Use master nodepool

jamielennox:
For now all we need to do is start up a nodepool server and point it at
it. The old state will need to be rebuilt but that's ok for us.


omgjlk (approved):
This looks fine to me. Maybe we should wait until the dust settles from re-opening Hoist before merging it in.

jamielennox:
I'm going to merge this so i have a bit of time to play with it.

anne-bonny:
Starting gate_github jobs.

PR #7, opened 2016-12-15
Remove the ssh-agent handling from .travis.yml

jamielennox:
The ssh agent handling was to allow forwarding the ssh keys that would
allow travis to deploy from a github enterprise account. This isn't need
and doesn't seem to be available on public travis workers.


rattboi (approved):
Makes sense :+1:

PR #9, opened 2016-12-16
Publish logs from echo-true job

jamielennox:
Just because it doesn't do anything doesn't mean we don't want to
publish logs from the echo-true job.


omgjlk (approved):
LGTM.

PR #11, opened 2016-12-16
Add hoist as a project

omgjlk:
Only echo-true for now, we'll add tests as we figure out how to do them
in zuul land


rattboi (approved):
LGTM

PR #15, opened 2016-12-19
Remove --no-images from nodepool commands

jamielennox:
--no-images seems to have been removed from nodepool and it prevents the
delete and launcher being started.


omgjlk:
So this failed CI because of a new requirement. https://github.com/BonnyCI/hoist/blob/master/CONTRIBUTING.md has more, but you'll need a signed-off-by line in your comment. The check itself should have returned failure, but it instead timed out on the check, weird.

omgjlk (dismissed):
signed-off-by

omgjlk (approved):

anne-bonny:
Starting gate_github jobs.

PR #16, opened 2016-12-19
WIP - Add a nodepool check to datadog

loafyloaf:
This will add a datadog check for the nodepool status page.

Signed-off-by: matthew wagoner <mwagone@us.ibm.com>


roles/dd-nodepool/files/nodepool_server.py
```diff
@@ -0,0 +1,47 @@
+import time
```
omgjlk, line 15: We likely need some licensing information here.

omgjlk (changes requested):
Needs some licensing, and probably need to squash these commits down into one, that has a signed-of-by line in the commit itself. Every commit has to have a signed-off-by. If this is just a work in progress, please tag the subject of the PR such to make it clear that it's not ready for review yet.

PR #18, opened 2016-12-20
Install os-loganalyze

jamielennox:
os-loganalyze is openstack-infra's log formatter that adds some basic
formatting around the raw logs.

My takeaway from this is when we are serious we can do our own one of
these better than os-loganalyze. It's clunky to install, not pretty and
very basicly done.

For now however we can use this to display logs faster than writing our
own formatter.


install-ci.yml
```diff
@@ -82,20 +82,88 @@
...
+            RewriteCond %{DOCUMENT_ROOT}%{REQUEST_FILENAME} !-l
+            RewriteCond %{REQUEST_FILENAME} !^/icon
+            RewriteRule ^/(.*)$ /htmlify/$1 [QSA,L,PT,NS]
+
+            WSGIDaemonProcess os-loganalyze processes=5 threads=1 display-name=%{GROUP}
+            WSGIScriptAlias /htmlify /etc/os_loganalyze/wsgi/os-loganalyze.py
+            WSGIApplicationGroup %{GLOBAL}
+
+            ServerSignature Off
+
```
omgjlk, line 166: This seems like a LOT of data to be passing in at role call, a lot of configuration in the playbook itself rather than a inventory var or role var.

roles/os-loganalyze/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+    - git
+
+- name: create dirs
+  file:
+    state: directory
+    path: "{{ item }}"
+    mode: 0755
+  with_items:
+    - /etc/os_loganalyze
+    - /etc/os_loganalyze/wsgi
```
omgjlk, line 18: You could just list the wsgi child path. All leading parent paths will be created with the provided details.

roles/os-loganalyze/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+
+- name: create config files
+  template:
+    src: "etc/os_loganalyze/{{ item }}"
+    dest: "/etc/os_loganalyze/{{ item }}"
+  with_items:
+    - wsgi.conf
+    - file_conditions.yaml
+    - wsgi/os-loganalyze.py
```
omgjlk, line 27: I think you forgot to git add this last file path.

omgjlk (dismissed):
CI failing:

```
failed: [allinone] (item=wsgi/os-loganalyze.py) => {"failed": true, "item": "wsgi/os-loganalyze.py", "msg": "Unable to find 'etc/os_loganalyze/wsgi/os-loganalyze.py' in expected paths."}
```

install-ci.yml
jamielennox: yep, i agree, but we haven't really got a precedent for this so i wasn't sure where to put it.

omgjlk (approved):

anne-bonny:
Starting gate_github jobs.

PR #20, opened 2016-12-21
Always update DIB

jamielennox:
We want to make sure that we are always running the latest
diskimage-builder because we are likely going to be making a lot of last
minute fixes there.


omgjlk (approved):
Seems reasonable.

anne-bonny:
Starting gate_github jobs.

PR #23, opened 2016-12-23
Configure the github status to link to the job logs

ConsoleCatzirl:
Depends-On: BonnyCI/zuul#6
Signed-off-by: K Jonathan Harker <jonathan.harker@ibm.com>

### DIFF
--- a/roles/zuul/templates/etc/zuul/config/layout.yaml
+++ b/roles/zuul/templates/etc/zuul/config/layout.yaml
@@ -21,10 +21,12 @@ pipelines:
     success:
       github:
         status: true
+        status_url: &log_link https://logs.bonnyci.com/{pipeline.name}/{change.project.name}/{change.number}/{item.enqueue_time}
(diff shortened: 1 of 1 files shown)


roles/zuul/templates/etc/zuul/config/layout.yaml
```diff
@@ -21,10 +21,12 @@ pipelines:
     success:
       github:
         status: true
+        status_url: &log_link https://logs.bonnyci.com/{pipeline.name}/{change.project.name}/{change.number}/{item.enqueue_time}
```
omgjlk, line 24: I feel like the "logs.bonnyci.com" bit needs to be a template value, if this were to be useable by other orgs for other URLs. Not a huge issue for now though.

omgjlk (commented):
Not approving yet as required change hasn't gone in.

jamielennox (approved):

PR #34, opened 2017-01-04
Enable sudo access to bonnyci user

jamielennox:
We may in future want to remove this but for now we need a way for the
CI user to install there own packages.


ConsoleCatzirl (approved):
LGTM

PR #36, opened 2017-01-05
Add ubuntu-minimal to the nodepool nodes

ConsoleCatzirl:
A cloud instance isn't much without networking.


jamielennox (approved):
I for one  like ubuntu-minimal. It's the basics of what i would expect to find on a "ubuntu" labelled image.

jamielennox:
I also don't think there's any need to be absolutely minimalist here. I think we want to provide a general purpose ubuntu node for people and i don't think the additional weight here is an issue.

jamielennox:
recheck

jamielennox:
recheck

jamielennox:
recheck

PR #55, opened 2017-01-11
Add ansible syntax check to hoist

ConsoleCatzirl:
Add an experimental pipeline triggered by "check experimental" comments,
and create an ansible-syntax job-template. Create a job from the template
for BonnyCI/hoist and put it in the experimental pipeline.

### DIFF
--- a/roles/zuul/files/jobs/ansible.yaml
+++ b/roles/zuul/files/jobs/ansible.yaml
@@ -1,0 +1,33 @@
+- builder:
+    name: install-ansible
+    builders:
+      - shell: |
+          echo Installing ansible into /opt/venvs/ansible
+          echo Be sure to source /opt/venvs/ansible/bin/activate in
+          echo your builder step to use this virtualenv
+
+          virtualenv /opt/venvs/ansible
+          source /opt/venvs/ansible/bin/activate
+          pip install ansible
+
+          if test -r requirements.txt; do
+            pip install -r requirements.txt
+          fi
+
+          if test -r requirements.yml; do
+            ansible-galaxy install -r requirements.yml
+          fi
+
+- job:
+    name: ansible-syntax
+    node: 'ubuntu-xenial'
+
+    builders:
+      - zuul-git-prep
+      - install-ansible
+      - shell: |
+          source /opt/venvs/ansible/bin/activate
+          ansible-playbook -i /dev/null --syntax-check $(find . tests -maxdepth 1 -type f -name \*.yml -not -name .travis.yml)
+
+    publishers:
+      - console-log
--- a/roles/zuul/files/jobs/hoist.yaml
+++ b/roles/zuul/files/jobs/hoist.yaml
@@ -4,13 +4,9 @@
 
     builders:
       - zuul-git-prep
+      - install-ansible
       - shell: |
-          tmpdir=$(mktemp -d)
-          trap "rm -rf $tmpdir" EXIT
-          virtualenv $tmpdir/venv
-          source $tmpdir/venv/bin/activate
-          pip install ansible
-          pip install -r requirements.txt
+          source /opt/venvs/ansible/bin/activate
           ansible-playbook -i inventory/nodepool.py install-ci.yml --skip-tags monitoring -e @secrets.yml.example
           ansible-playbook -i inventory/nodepool.py  tests/validate-ci.yml
 
--- a/roles/zuul/templates/etc/zuul/config/layout.yaml
+++ b/roles/zuul/templates/etc/zuul/config/layout.yaml
@@ -65,6 +65,7 @@ projects:
   - name: BonnyCI/hoist
     check_github:
       - ubuntu-hoist
+      - ansible-syntax
     gate_github:
       - echo-true
 


anne-bonny:
Build failed.

- http://logs.bonnyci.com/check_github/BonnyCI/hoist/55/1484176367.44/ubuntu-hoist : FAILURE in 3m 49s

roles/zuul/files/jobs/ansible.yaml
```diff
@@ -0,0 +1,30 @@
+- builder:
+    name: install-ansible
+    builders:
+      - shell: |
+          tmpdir=$(mktemp -d)
+          trap "rm -rf $tmpdir" EXIT
+          virtualenv $tmpdir/venv
+          source $tmpdir/venv/bin/activate
+          pip install ansible
```
jamielennox, line 11: i would think to use this as a builder you don't want to do the trapdir, but go for /opt/venvs/ansible and then symlink at least ansible, ansible-playbook and ansible-galaxy into /usr/local/bin or somewhere.

roles/zuul/files/jobs/hoist.yaml
```diff
@@ -4,15 +4,16 @@
...
           ansible-playbook -i inventory/nodepool.py  tests/validate-ci.yml
 
     publishers:
       - console-log
+
+- project:
+    name: BonnyCI/hoist
+    jobs:
+      - ansible-syntax-{name}:
+        files: "bastion.yml install-ci.yml provision.yml tests/validate-ci.yml"
```
jamielennox: nit: from a yaml perspective i would expect files to be a list here - not sure if it's possible to flatten a list into space seperated variables.

roles/zuul/templates/etc/zuul/config/layout.yaml
```diff
@@ -67,6 +81,8 @@ projects:
       - ubuntu-hoist
     gate_github:
       - echo-true
+    experimental_github:
+      - ansible-syntax-BonnyCI/hoist
```
jamielennox: why put syntax check in experimental?

jamielennox (dismissed):
Main comment, i don't think the builder will work like that with the tmp virtualenv

roles/zuul/templates/etc/zuul/config/layout.yaml
jamielennox: as opposed to just check

roles/zuul/files/jobs/hoist.yaml
ConsoleCatzirl: Yeah, it would be nice to use a yaml sequence here, but these values are simply passed through to format() in python, and I don't know if we can change how that's interpreted.

anne-bonny:
Build failed.

- http://logs.bonnyci.com/check_github/BonnyCI/hoist/55/1484182433.09/ubuntu-hoist : FAILURE in 4m 19s

roles/zuul/templates/etc/zuul/config/layout.yaml
ConsoleCatzirl: because I'm not sure if it works, and I figured we would eventually want to add an experimental pipeline. but i suppose it's overkill for this job since our check pipeline isn't voting yet

roles/zuul/files/jobs/ansible.yaml
```diff
@@ -0,0 +1,33 @@
...
+          virtualenv /opt/venvs/ansible
+          source /opt/venvs/ansible/bin/activate
+          pip install ansible
+
+          if test -r requirements.txt; do
+            pip install -r requirements.txt
+          fi
+
+          if test -r requirements.yml; do
+            ansible-galaxy install -r requirements.txt
```
nibalizer: I thought this file was called roles.yml usually?

roles/zuul/files/jobs/ansible.yaml
```diff
@@ -0,0 +1,33 @@
...
+- job-template:
+    name: ansible-syntax-{name}
+    node: 'ubuntu-xenial'
+
+    builders:
+      - zuul-git-prep
+      - install-ansible
+      - shell: |
+          source /opt/venvs/ansible/bin/activate
+          ansible-playbook -i /dev/null --syntax-check {files}
```
nibalizer: Why not just ``find . -type -f -name '*.yml'`` ? The YAML spec says the file extension is .yaml so .yml should only be ansilbe.

roles/zuul/templates/etc/zuul/config/layout.yaml
nibalizer: I agree with jamie that this can go straight into check. I agree with jesuasurus that we will/do need an experimental pipeline.

nibalizer (commented):
Looks good, feel free to merge once the test is put in the check queue. I have a couple questions that might make the code better, but you don't have to take those suggesetions if you don't want to.

roles/zuul/files/jobs/ansible.yaml
ConsoleCatzirl: We only want to check the playbooks, not the tasks or defaults. But adding a `-depth=1` to the find would work for that.

roles/zuul/files/jobs/ansible.yaml
ConsoleCatzirl: The galaxy docs refer to requirements.yml

roles/zuul/files/jobs/ansible.yaml
```diff
@@ -0,0 +1,33 @@
...
+- job:
+    name: ansible-syntax
+    node: 'ubuntu-xenial'
+
+    builders:
+      - zuul-git-prep
+      - install-ansible
+      - shell: |
+          source /opt/venvs/ansible/bin/activate
+          ansible-playbook -i /dev/null --syntax-check $(find . tests -depth 1 -name \*.yml)
```
nibalizer: Probably want -type f (just because people are terrible)
Why is there a 'find . tests' ? Wouldn't the tests dir be inside the . scope?

ConsoleCatzirl: but we only want to look at a depth of 1, so we won't match anything under tests/ unless we add that as a base path

nibalizer (approved):

PR #62, opened 2017-01-18
Add log rotation

ConsoleCatzirl:
Add logrotate role and configure rotation of zuul logs.

---
This change is [Reviewable](https://reviewable.io/reviews/bonnyci/hoist/62)

### DIFF
--- a/roles/logrotate/tasks/main.yml
+++ b/roles/logrotate/tasks/main.yml
@@ -1,0 +1,12 @@
+---
+- name: Install base package
+  apt:
+    name: logrotate
+    state: installed
+
+- name: Install config
+  template:
+    src: etc/logrotate.d/template.conf
+    dest: /etc/logrotate.d/{{ item.name }}
+  with_items: "{{ logrotate_configs }}"
+  when: logrotate_configs|length > 0
--- a/roles/logrotate/tasks/main.yml
+++ b/roles/logrotate/tasks/main.yml
@@ -9,4 +9,3 @@
     src: etc/logrotate.d/template.conf
     dest: /etc/logrotate.d/{{ item.name }}
   with_items: "{{ logrotate_configs }}"
-  when: logrotate_configs|length > 0
--- a/install-ci.yml
+++ b/install-ci.yml
@@ -55,6 +55,17 @@
       tags:
         - monitoring
 
+    - role: logrotate
+      logrotate_configs:
+        - name: zuul
+          path: /var/log/zuul/*log
+          options:
+            - compress
+            - missingok
+            - rotate 30
+            - daily
+            - notifempty
+
 - name: Install nodepool
   hosts: nodepool
   become: yes
--- a/roles/logrotate/defaults/main.yml
+++ b/roles/logrotate/defaults/main.yml
@@ -1,0 +1,1 @@
+logrotate_configs: []
--- a/roles/logrotate/templates/etc/logrotate.d/template.conf
+++ b/roles/logrotate/templates/etc/logrotate.d/template.conf
@@ -1,0 +1,5 @@
+"{{ item.path }}" {
+  {% for option in item.options -%}
+  {{ option }}
+  {% endfor -%}
+}


roles/logrotate/templates/etc/logrotate.d/template.conf.j2
```diff
@@ -0,0 +1,5 @@
+"{{ item.path }}" {
```
rattboi: Pretty sure we have a policy of not putting .j2 on our templates.

roles/zuul/meta/main.yml
```diff
@@ -1,2 +1,12 @@
 dependencies:
   - pip
+  - role: logrotate
+    logrotate_configs:
```
rattboi: Not positive how I feel about having the config values in meta. That being said, maybe that's just my lack of experience.

rattboi (dismissed):
One change request (template filename), currently, but also wanna discuss keeping logrotate config in meta/ (not sure if it can be in defaults or not).

roles/logrotate/templates/etc/logrotate.d/template.conf.j2
ConsoleCatzirl: I hadn't noticed, I can remove that from the filename.

roles/zuul/meta/main.yml
ConsoleCatzirl: The idea is to reuse the same role on different hosts, each would have a unique list of files to rotate.

rattboi: Can meta source things in a zuul's role's defaults? I prefer to have all the config values in one place, if possible.

ConsoleCatzirl: I don't think one role has access to another role's defaults, but I can test that.

ConsoleCatzirl: Putting the dict in the zuul role's defaults doesn't work, but putting it in the zuul group_vars does.

eggshell: iirc role defaults' scope only goes as far as the role they live in, and are lowest priority. The caveats to this would be role dependencies and the `include_role` module in 2.2+

http://docs.ansible.com/ansible/playbooks_roles.html#role-default-variables
http://docs.ansible.com/ansible/include_role_module.html

I agree that role defaults should be in `defaults/main.yml`. If they need to be accessed in a different role then `include_role` or role dependencies in `meta` can be exercised. I don't have immense experience with either, so making a recommendation on which one to go with right now is difficult.

inventory/group_vars/zuul
```diff
@@ -17,3 +17,13 @@ bonnyci_zuul_apache_vhosts:
...
+
+logrotate_configs:
+  - name: zuul
+    path: /var/log/zuul/*log
+    options:
+      - compress
+      - missingok
+      - rotate 30
+      - daily
+      - notifempty
```
omgjlk: I think this should be defined in zuul role defaults. Those can be used at role dependency time.

roles/logrotate/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+  apt:
+    name: logrotate
+    state: installed
+
+- name: Install config
+  template:
+    src: etc/logrotate.d/template.conf
+    dest: /etc/logrotate.d/{{ item.name }}
+  with_items: "{{ logrotate_configs }}"
+  when: logrotate_configs|length > 0
```
omgjlk: Shouldn't the service be ensured enabled/running? And does the service need to be touched in some way to pick up new/changing config?

roles/zuul/meta/main.yml
```diff
@@ -1,2 +1,3 @@
 dependencies:
   - pip
+  - logrotate
```
omgjlk: If we have the log rotation data as a role default, you can then do

```
dependencies:
  - role: pip
  - role: logrotate
    logrotate_configs: "{{ zuul.logrotate_configs }}"
```

This way the info about the logging stays with the service, and is only altered if some site REALLY needs to via inventory vars. When it's all in inventory it becomes tedious to create new inventories.

omgjlk (dismissed):
Needs a bit more work.

inventory/group_vars/zuul
jesusaurus: I tested that, and the dict was not picked up.

roles/logrotate/tasks/main.yml
ConsoleCatzirl: Some distros use systemd directly to run logrotate, but it looks like xenial still uses cron to run logrotate.

roles/zuul/meta/main.yml
ConsoleCatzirl: Why is it better to have this extra abstraction, rather than just put the dict in the meta file?

omgjlk: When it's directly in the meta file, it cannot be overridden by other deployers, if we have any other deployers. I'm not sure if we want this to be a configurable thing within Hoist, or if it should be static.

Also, I prefer as much of the role configuration to be in the defaults, rather than scattered in other files.

jamielennox: So i disagree here - but i defer to experience. 

IMO roles should define the bare minimum of things that you use to configure that piece of the stack. In this case the zuul role should do nothing but configure zuul. You can then define on a per host/group level that i want to have zuul here, and also logrotate that looks for output configured by the zuul piece.

I don't know how likely it is someone will reuse our roles but i'd ideally like to be able to publish a standalone zuul role that could be used by infra or someone else and I don't think it should be a requirement of setting up zuul that you have to have logrotate, or a filebeat, or whatever else we log with. That's not part of a part of setting up zuul - IMO. 

The practical side of thi is that i don't think logrotate should be a dependency here, it should be listed in the roles: for hosts: zuul in install-ci.yml

roles/logrotate/tasks/main.yml
jamielennox: nit: i think when: is redunant here, if length == 0 then the with_items won't run.

roles/zuul/meta/main.yml
rattboi: What jamie is saying makes sense to me. We should probably avoid unnecessary dependencies if we can help it.

roles/logrotate/tasks/main.yml
omgjlk: That's a correct nit. If `logrotate_configs` is an empty list then the task won't fire.

roles/zuul/meta/main.yml
omgjlk: I can see Jamie's point of view too. Mostly I want to avoid this data being defined solely in inventory files. We have to either be clever with some default inventory variables (a shared "all" group), implement a plugin to express default values for all inventories (like Ursula has), or remember to copy/paste/update the data into every inventory.

Perhaps we can have some roles that are meant to be shared, like zuul, and some roles that are our implementation specific, like the log rotate role. The data for logs to rotate may have to be duplicated, as some of the info may exist in the zuul role, but if we don't want to establish a relationship we'll have to duplicate it.

jamielennox (approved):

omgjlk (approved):

anne-bonny:
Starting gate_github jobs.

PR #64, opened 2017-01-18
Fix signed-off-by-test.sh for best practices

mlangbehn:
Script was updated in anticipation of `shellcheck` being added to CI
tests.

Signed-off-by: Matt Langbehn <matthew.langbehn@gmail.com>

PR #67, opened 2017-01-18
Add shellcheck to lint shell scripts

mlangbehn:
Now, all shell scripts are linted for best practices.

Requires: #64, #66, #70, and #71

### DIFF
--- a/roles/zuul/files/jobs/hoist.yaml
+++ b/roles/zuul/files/jobs/hoist.yaml
@@ -1,3 +1,4 @@
+---
 - job:
     name: ubuntu-hoist
     node: 'ubuntu-hoist'
@@ -20,5 +21,6 @@
     builders:
       - zuul-git-prep
       - shell: |
-          ./tests/signed-off-by-test.sh
           python tests/layout-checks.py job-list.txt
+          ./tests/shellcheck-test.sh
+          ./tests/signed-off-by-test.sh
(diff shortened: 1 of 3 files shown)


roles/zuul/files/jobs/hoist.yaml
```diff
@@ -22,3 +23,4 @@
       - shell: |
           ./tests/signed-off-by-test.sh
           python tests/layout-checks.py job-list.txt
+          ./tests/shellcheck-test.sh
```
eggshell, line 25: This is a nit to the nth degree, but can we keep tests of the same file extension grouped together like they are in `.travis.yml`?

eggshell (changes requested):
One nit inline. Other than that we just need #66 to merge first.

roles/zuul/files/jobs/hoist.yaml
mlangbehn: We certainly can. I added this test to the end because I'm not entirely sure how this mechanism works. In Travis CI, we are running the tests in parallel, but I think they happen sequentially here. If they are sequential, I think it makes sense to add this check to the end?

eggshell: They happen in parallel iirc.

Similar to: https://blog.travis-ci.com/2012-11-28-speeding-up-your-tests-by-parallelizing-them/

mlangbehn: If they were happening in parallel, wouldn't each line end with an `&`?

eggshell: I'm not sure why that would be needed since it seems to take each line as a separate thing to run and it goes off and runs each one (all starting at the same time).

mlangbehn: Just rearranged all three to be in alphabetical order.

eggshell (approved):
lgtm once #66 merges and this passes CI fully.

rattboi:
recheck

anne-bonny:
Build failed.

- http://logs.bonnyci.com/check_github/BonnyCI/hoist/67/1485194029.72/ubuntu-hoist : FAILURE in 23s
- http://logs.bonnyci.com/check_github/BonnyCI/hoist/67/1485194029.72/ansible-syntax : SUCCESS in 53s
- http://logs.bonnyci.com/check_github/BonnyCI/hoist/67/1485194029.72/hoist-checks : SUCCESS in 8s

mlangbehn:
recheck

PR #71, opened 2017-01-23
Fix vagrant-inject-pubkey.sh for best practices

mlangbehn:
Script was updated in anticipation of `shellcheck` being added to CI
tests.


rattboi (approved):
Although I prefer the dataflow style, I don't feel strongly enough to make a special case for it for shellcheck.

PR #79, opened 2017-01-27
Add bonnyci-run-pipeline job for executing in-repo test scripts

gandelman-a:
This adds a job template that can be re-used to run custom in-repo
job scripts for specified pipelines.

The job indicates to the script the pipeline in which the job is being run
via the 'BONNYCI_TEST_PIPELINE' environment variable.  It also exposes
a local directory via 'BONNYCI_TEST_LOG_DIR' which users may fill with
test output to be archived by the new 'bonnyci-logs' publisher.

The new job has been added to the sandbox repo for experimenting.

### DIFF
--- a/roles/nodepool/files/etc/nodepool/elements/bonnyci-nodepool/install.d/90-test-logs-dir
+++ b/roles/nodepool/files/etc/nodepool/elements/bonnyci-nodepool/install.d/90-test-logs-dir
@@ -1,5 +1,4 @@
 #!/bin/bash
-# Copyright (C) 2011-2013 OpenStack Foundation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -21,6 +20,8 @@ if [ "${DIB_DEBUG_TRACE:-0}" -gt 0 ]; then
 fi
 set -e
 
-# Our use requires zuul be installed so that we can call zuul-cloner
-# The source is put in place by the source-repositories element.
-pip2 install /opt/zuul/
+# Create a log directory for tests to stuff things to be archived by publisher.
+# This path is also set in the job definition and exported to in-repo test
+# scripts, at ./roles/zuul/files/jobs/bonnyci.yaml.
+mkdir -p /opt/bonnyci/logs
+chown bonnyci:bonnyci /opt/bonnyci/logs
--- a/roles/zuul/files/jobs/bonnyci.yaml
+++ b/roles/zuul/files/jobs/bonnyci.yaml
@@ -1,0 +1,49 @@
+- publisher:
+    name: bonnyci-logs
+    publishers:
+      - scp:
+          site: bonnyci-scp
+          files:
+            - target: 'logs/$LOG_PATH'
+              source: '/opt/bonnyci/logs/**'
+              keep-hierarchy: true
+              copy-after-failure: true
+
+- job-template:
+    name: 'bonnyci-run-{pipeline}'
+    id: 'bonnyci-run-pipeline-{pipeline}'
+    test_log_dir: /opt/bonnyci/logs/
+    builders:
+      - zuul-git-prep
+      - shell: |
+            #!/bin/bash -xe
+            if [[ ! -f .bonnyci/run.sh ]]; then
+                echo "ERROR: No test script found at .bonnyci/run.sh, skipping"
+                exit 1
+            elif [[ ! -x .bonnyci/run.sh ]]; then
+                echo "ERROR: Test script found at .bonnyci/run.sh but it is not executable"
+                exit 1
+            fi
+
+            export BONNYCI_TEST_PIPELINE={pipeline}
+            export BONNYCI_TEST_LOG_DIR={test_log_dir}
+
+            if ./.bonnyci/run.sh; then
+                echo ".bonnyci/run.sh test(s) passed :)"
+                exit 0
+            else
+                echo ".bonnyci/run.sh test(s) failed :("
+                exit 1
+            fi
+
+    publishers:
+      - console-log
+      - bonnyci-logs
+
+- project:
+    name: bonnyci-run-pipeline
+    jobs:
+        - 'bonnyci-run-pipeline-{pipeline}':
+            pipeline: check
+        - 'bonnyci-run-pipeline-{pipeline}':
+            pipeline: gate
--- a/roles/zuul/templates/etc/zuul/config/layout.yaml
+++ b/roles/zuul/templates/etc/zuul/config/layout.yaml
@@ -74,12 +74,9 @@ projects:
 
   - name: BonnyCI/sandbox
     check_github:
-      - echo-true
-      - noop
+      - bonnyci-run-check
     gate_github:
-      - echo-true
-      - noop
-
+      - bonnyci-run-gate
   - name: BonnyCI/zuul
     check_github:
       - tox_py27
--- a/roles/nodepool/files/etc/nodepool/elements/bonnyci-nodepool/element-deps
+++ b/roles/nodepool/files/etc/nodepool/elements/bonnyci-nodepool/element-deps
@@ -1,2 +1,3 @@
+devuser
 package-installs
 source-repositories


anne-bonny:
Build failed.

- http://logs.bonnyci.com/check_github/BonnyCI/hoist/79/1485553476.72/ubuntu-hoist : SUCCESS in 4m 41s
- http://logs.bonnyci.com/check_github/BonnyCI/hoist/79/1485553476.72/ansible-syntax : SUCCESS in 55s
- http://logs.bonnyci.com/check_github/BonnyCI/hoist/79/1485553476.72/hoist-checks : FAILURE in 14s

roles/zuul/files/jobs/bonnyci.yaml
```diff
@@ -0,0 +1,43 @@
...
+            export BONNYCI_TEST_LOG_DIR={test_log_dir}
+
+            (cd .bonnyci/ && ./run.sh)
+
+            if [[ $? -eq 0 ]]; then
+                echo ".bonnyci/run.sh test(s) passed :)"
+                exit 0
+            else
+                echo ".bonnyci/run.sh test(s) failed :("
+                exit 0
```
ConsoleCatzirl: This should exit 1

roles/zuul/templates/etc/zuul/config/layout.yaml
```diff
@@ -74,12 +74,11 @@ projects:
 
   - name: BonnyCI/sandbox
     check_github:
-      - echo-true
-      - noop
+      - 'bonnyci-run-{pipeline}':
```
ConsoleCatzirl: We also need to create a job from the job-template. We don't currently have any job-templates, so we will need to create `roles/zuul/files/jobs/projects.yaml` similar to [infra's projects.yaml](http://git.openstack.org/cgit/openstack-infra/project-config/tree/jenkins/jobs/projects.yaml) with the job-template and parameters in the jobs section of a BonnyCI/sandbox project.

roles/zuul/files/jobs/bonnyci.yaml
```diff
@@ -0,0 +1,43 @@
...
+            if [[ ! -f .bonnyci/run.sh ]]; then
+                echo "ERROR: No test script found at .bonnyci/run.sh, skipping"
+                exit 1
+            elif [[ ! -x .bonnyci/run.sh ]]; then
+                echo "ERROR: Test script found at .bonnyci/run.sh but it is not executable"
+                exit 1
+            fi
+
+            export BONNYCI_TEST_PIPELINE={pipeline}
+            export BONNYCI_TEST_LOG_DIR={test_log_dir}
```
ConsoleCatzirl, line 29: If we set these in bonnyci_functions.py then we can use BONNYCI_TEST_LOG_DIR in the publisher source and won't need to pass in a parameter to set it, reducing the number of places this is hard-coded and making it easier to update. But since this is only meant to be a temporary solution I don't feel very strongly about making it easy to update.

roles/nodepool/files/etc/nodepool/elements/bonnyci-nodepool/install.d/90-test-logs-dir
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Copyright (C) 2011-2013 OpenStack Foundation
```
ConsoleCatzirl: In this correct?

roles/zuul/files/jobs/bonnyci.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+            (cd .bonnyci/ && ./run.sh)
```
ConsoleCatzirl: Why change the working directory? I think it makes sense for the tests to run from the root of the repository, but that might be an argument to not put the entrypoint in a subdirectory.

ConsoleCatzirl (changes requested):

roles/zuul/files/jobs/bonnyci.yaml
gandelman-a: The way I envision people using this (and I could be wrong!), would be to have .bonnyci be a self-contained thing, probably with a few helper libraries or scripts in there that the main run.sh would source or call.  By executing run.sh within the subdir, the main script can just assume the things it needs are accessible via a path relative to $0.  I kinda had devstack plugins in the back of my head when putting this together.  Happy to hear what others think.

roles/nodepool/files/etc/nodepool/elements/bonnyci-nodepool/install.d/90-test-logs-dir
```diff
@@ -0,0 +1,27 @@
...
+if [ ${DIB_DEBUG_TRACE:-0} -gt 0 ]; then
+    set -x
+fi
+set -e
+
+# Create a log directory for tests to stuff things to be archived by publisher.
+# This path is also set in the job definition and exported to in-repo test
+# scripts, at ./roles/zuul/files/jobs/bonnyci.yaml.
+mkdir -p /opt/bonnyci/logs
+chown bonnyci:bonnyci /opt/bonnyci/logs
```
jamielennox, line 27: Why not just ~/logs?

roles/zuul/files/jobs/bonnyci.yaml
jamielennox: setting it in bonnyci_functions.py would make it available for all tests. I'm not opposed to that but if we expect these to only be defined for this job type then it doesn't really matter.

roles/zuul/files/jobs/bonnyci.yaml
jamielennox: I think i would expect to be in the root directory of the code checkout. That root directory is not as predeictable as we might like, it's something like ~/workspace/{pipeline}/{repo} so it's not easy to go back to that dir to run something like tox. 

On the other hand if you make .bonnyci a well known folder location it's super simple for people in scripts to do source .bonnyci/abc.sh. 

I would vote for remaining in the checkout dir.

gandelman-a: OK, executing from the root of the checkout it is.

roles/nodepool/files/etc/nodepool/elements/bonnyci-nodepool/install.d/90-test-logs-dir
gandelman-a: We'll be referencing this path on the zuul config side, in the job's publisher. I'd rather it be a static full path that is indepenent of the user's home dir, which could change for some reason in the future.

roles/nodepool/files/etc/nodepool/elements/bonnyci-nodepool/install.d/90-test-logs-dir
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied.
+#
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# dib-lint: disable=setu setpipefail
+if [ ${DIB_DEBUG_TRACE:-0} -gt 0 ]; then
```
mlangbehn: This should be either:
```shell
if [ "${DIB_DEBUG_TRACE:-0}" -gt 0 ]; then
```
or:
```shell
if ${DIB_DEBUG_TRACE:-0}; then
```

roles/zuul/files/jobs/bonnyci.yaml
```diff
@@ -0,0 +1,51 @@
...
+              copy-after-failure: true
+
+- job-template:
+    name: 'bonnyci-run-{pipeline}'
+    id: 'bonnyci-run-pipeline-{pipeline}'
+    test_log_dir: /opt/bonnyci/logs/
+    builders:
+      - zuul-git-prep
+      - shell: |
+            #!/bin/bash
```
mlangbehn:
```shell
#!/bin/bash -ex
```

roles/zuul/files/jobs/bonnyci.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+            ./.bonnyci/run.sh
+
+            if [[ $? -eq 0 ]]; then
+                echo ".bonnyci/run.sh test(s) passed :)"
+                exit 0
+            else
+                echo ".bonnyci/run.sh test(s) failed :("
+                exit 1
+            fi
```
mlangbehn, line 37: This block can be simplified to:
```shell
if ./.bonnyci/run.sh; then
    echo ".bonnyci/run.sh test(s) passed :)"
    exit 0
else
    echo ".bonnyci/run.sh test(s) failed :("
    exit 1
fi
```

mlangbehn (changes requested):

roles/nodepool/files/etc/nodepool/elements/bonnyci-nodepool/install.d/90-test-logs-dir
gandelman-a: OK, I think this is just boilerplated out from other DIB elements. Copied it from 90-zuul-cloner, updated there as well.

roles/zuul/files/jobs/bonnyci.yaml
gandelman-a: done

omgjlk (approved):

mlangbehn (approved):
Looks good, thanks!

roles/zuul/files/jobs/bonnyci.yaml
mlangbehn: Thanks.

ConsoleCatzirl (approved):

anne-bonny:
Starting gate_github jobs.

PR #81, opened 2017-01-31
Manage datadog monitors from file

jamielennox:
So datadog-builder is basically jenkins-job-builder for datadog. It's
kind of ruff (pun intended) but creates and deletes monitors on datadog
based on the entries in a config file.

This is the bastion equivalent that is checking out and running the jobs
on a cron job. The idea in future would be to put this on a CI job or
something so monitors are updated whenever the config file is changed.

I know it is basically what ansible gives us, but it will give us a bit
more control over the tags, monitor locking and whatever else.

### DIFF
--- a/bastion.yml
+++ b/bastion.yml
@@ -15,6 +15,11 @@
         - monitoring
       when: secrets is defined
     - bastion
+    - role: datadog-builder
+      datadog_builder_secrets:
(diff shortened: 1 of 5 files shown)


rattboi (approved):
LGTM

PR #88, opened 2017-01-31
Remove deprecated nodepool argument

ConsoleCatzirl:
As of commit openstack-infra/nodepool@1e627a9 the nodepool argument `--no-builder` is deprecated,
and a separate builder process is required. Remove the argument from the
service files as suggested by the deprecation warning logged in the
nodepool log files.


mlangbehn (approved):
Looks good to me.

PR #91, opened 2017-02-02
Ensure node/npm are installed for lore testing

eggshell:
lore's markdownlint and textlint tests rely on npm to install their
dependencies.

Relevant failure:
http://logs.bonnyci.com/check_github/BonnyCI/lore/48/1485995068.93/lore/console.html

Signed-off-by: Cullen Taylor <CullenTaylor@outlook.com>

### DIFF
--- a/roles/zuul/files/jobs/lore.yaml
+++ b/roles/zuul/files/jobs/lore.yaml
@@ -6,6 +6,10 @@
       - zuul-git-prep
       - shell: |
           #!/bin/bash -ex
+          sudo apt-get install -yqq curl
+          curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.0/install.sh | bash
+          source ~/.bashrc
+          nvm install stable && nvm use stable
           ./tests/markdownlint-cli-test.sh
           ./tests/textlint-test.sh
           ./tests/shellcheck-test.sh


roles/zuul/files/jobs/lore.yaml
```diff
@@ -6,6 +6,7 @@
       - zuul-git-prep
       - shell: |
           #!/bin/bash -ex
+          sudo apt-get install -yqq nodejs npm
```
rattboi: Do we know what version of nodejs we get this way? Maybe it makes more sense to use nvm and pin?

eggshell: good call. We have travis set to `nvm install stable` so I'll do that here.

roles/zuul/files/jobs/lore.yaml
```diff
@@ -6,6 +6,10 @@
       - zuul-git-prep
       - shell: |
           #!/bin/bash -ex
+          sudo apt-get install -yqq curl
+          curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.0/install.sh | bash
```
mlangbehn, line 10: Would it make more sense to do this in dib?

eggshell: I don't think preloading all of our images with node is the greatest of ideas.

mlangbehn: You don't think our `ubuntu-xenial` image should have node? I'm not worried about performance tuning our builds yet, but I do think it would be a good idea to have some basic tools available.

I'm fine with this for now, but we will need to have images preloaded with software for projects to install their dependencies at some point.

eggshell: Right, but that's definitely a larger conversation. The point of our images (from what I can tell) is to be as bare as possible to test that your dependency installation is robust. Later on I'm sure we'll be adding images for different languages and toolsets, but node seems non-essential for our base image since it's only being used for linting one project.

mlangbehn (approved):

rattboi:
gate-recheck

anne-bonny:
Starting gate_github jobs.

PR #104, opened 2017-02-03
Make sure unwanted zuul services are stopped

ConsoleCatzirl:

### DIFF
--- a/roles/zuul/tasks/main.yml
+++ b/roles/zuul/tasks/main.yml
@@ -136,6 +136,19 @@
   include: merger.yml
   when: "'zuul-merger' in zuul_components"
 
+- name: Stop unwanted zuul services
+  service:
+    name: "{{ item }}"
+    state: stopped
+    enabled: no
+  when: "item not in zuul_components"
+  with_items:
+    - zuul-launcher
+    - zuul-merger
+    - zuul-server
+  register: stop_zuul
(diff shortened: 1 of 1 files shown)


roles/zuul/tasks/main.yml
```diff
@@ -136,6 +136,17 @@
...
+- name: Stop unwanted zuul services
+  service:
+    name: "{{ item }}"
+    state: stopped
+    enabled: no
+  when: "item not in zuul_components"
+  with_items:
+    - zuul-launcher
+    - zuul-merger
+    - zuul-server
```
omgjlk, line 148: My concern with this is that this module will throw an error if the service doesn't exist on the box. At Blue Box we had a version of the module that took an extra option that ignored the service being missing. I'd rather not carry that over here, but instead construct a failed_when setup that handles the scenario.

omgjlk (dismissed):

roles/zuul/tasks/main.yml
jamielennox: I was thinking more what about in the like all-in-one case where zuul_components may be specified on a role or something and executed multiple times. does that still work?

ConsoleCatzirl: If we were to switch to multiple zuul roles, this would probably fall apart. But currently zuul_components is going to be constant on each host.

omgjlk (approved):

PR #107, opened 2017-02-04
Update zuul group_vars for split-out merger

ConsoleCatzirl:
Now that the merger runs on a separate host, remove the component from
the zuul group vars. There are host vars in place for zuul.multinode
and allinone so that the tests will still put all three components on the
same host.


omgjlk (approved):
Looks good!

PR #108, opened 2017-02-04
Set zuul_merger_url based on ansible_nodename

ConsoleCatzirl:
Since ansible and nodepool nodes can reach mergers using the same host name,
automatically build the url based on the full hostname ansible uses to reach
the mergers.

If we ever have to change the url that nodepool nodes reach the mergers at
(e.g. if we add another nodepool provider), we may need to set the merger
url in host_vars for every merger.


ConsoleCatzirl:
recheck

PR #111, opened 2017-02-05
Remove zuul-merger group_vars from zuul

jamielennox:
When moving the zuul-merger over to its own host we left the group_vars
on the zuul group. We shouldn't need them any more, remove it.


ConsoleCatzirl (approved):

anne-bonny:
Starting gate_github jobs.

PR #112, opened 2017-02-06
Move git repos to a different port

jamielennox:
Having the git repos on port 80 conflicts with the status page and
probably other things in future. This URL is all handled via variables
so the port is irrelevant, move it to something arbitrary.

### DIFF
--- a/install-ci.yml
+++ b/install-ci.yml
@@ -71,8 +71,8 @@
       zuul_merger_git_dir: "{{ bonnyci_zuul_merger_git_dir }}"
 
     - role: apache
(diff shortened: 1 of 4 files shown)


ConsoleCatzirl:
You will also need to update zuul_merger_url in inventory/group_vars/multinode to add the port number, since the multinode test needs to use IPs instead of hostnames.

jamielennox:
@jesusaurus good spot, thanks

ConsoleCatzirl (approved):

anne-bonny:
Starting gate_github jobs.

PR #115, opened 2017-02-07
Create logfile for datadog-builder cronjob

rattboi:
Currently, this cronjob attempts to write to cron log, but since we
don't have an MTA installed, that output goes nowhere. Having the
error output go to a specific log is what we are doing for our other
cronjobs, so I think this makes sense.

Signed-off-by: Bradon Kanyid <rattboi@gmail.com>

### DIFF
--- a/roles/datadog-builder/tasks/datadog-builder.yml
+++ b/roles/datadog-builder/tasks/datadog-builder.yml
@@ -20,11 +20,25 @@
     - /opt/venvs
     - /etc/datadog-builder
 
+- name: create datadog-builder group
+  group:
+    name: datadog-builder
+    state: present
+
 - name: create datadog-builder user
   user:
     name: datadog-builder
+    group: datadog-builder
     home: /var/lib/datadog-builder
     state: present
+
+- name: Create log file
+  file:
+    path: /var/log/datadog-builder
+    state: directory
+    mode: 0755
+    owner: datadog-builder
+    group: datadog-builder
 
 - name: Create datadog-builder secrets
   copy:
@@ -64,3 +78,4 @@
       /opt/venvs/datadog-builder/bin/datadog-builder
       --config /etc/datadog-builder/datadog-secrets.yml
       update {{ datadog_builder_monitor_dir }}/{{ datadog_builder_monitor_file }}
+      2>&1 >> /var/log/datadog-builder/datadog-builder.log


roles/datadog-builder/tasks/datadog-builder.yml
```diff
@@ -23,9 +23,18 @@
...
     home: /var/lib/datadog-builder
     state: present
 
+- name: Create log file
+  file:
+    path: /var/log/datadog-builder.log
+    state: present
+    mode: 0755
+    owner: datadog-builder
+    group: datadog-builder
```
jamielennox: I'm not sure if this is a nit or not. 

Normally the way to do this is create a folder in /var/log that has permissions that the application has rights that it can create new files in. This means it shouldn't get into trouble when we do things like logrotation or whatever.

Probably this would be ok, but i think the folder would be better.

roles/datadog-builder/tasks/datadog-builder.yml
```diff
@@ -64,3 +73,4 @@
       /opt/venvs/datadog-builder/bin/datadog-builder
       --config /etc/datadog-builder/datadog-secrets.yml
       update {{ datadog_builder_monitor_dir }}/{{ datadog_builder_monitor_file }}
+      2>> /var/log/datadog-builder.log > /dev/null
```
jamielennox: why is stdout going to /dev/null? is it not useful? 

If it's not we can do things like the standard python logging configuration in datadog-builder and make this more sane.

jamielennox (changes requested):

roles/datadog-builder/tasks/datadog-builder.yml
rattboi: Yeah, all log output seems to be on stderr

roles/datadog-builder/tasks/datadog-builder.yml
rattboi: Fixed

roles/datadog-builder/tasks/datadog-builder.yml
rattboi: I redirect stderr to stdout now, just in case we start getting useful things from stdout in the future

jamielennox (approved):

jamielennox:
gate_recheck

jamielennox (approved):

roles/datadog-builder/tasks/datadog-builder.yml
```diff
@@ -20,12 +20,26 @@
...
 
+- name: create datadog-builder group
+  group:
+    name: datadog-builder
+    state: present
+
 - name: create datadog-builder user
   user:
     name: datadog-builder
+    group: datadog-builder
```
ConsoleCatzirl, line 31: TIL ansible doesn't create a primary group with the same name as the user when creating a user.

roles/datadog-builder/tasks/datadog-builder.yml
```diff
@@ -20,12 +20,26 @@
...
     home: /var/lib/datadog-builder
     state: present
 
+- name: Create log file
+  file:
+    path: /var/log/datadog-builder
+    state: directory
+    mode: 0755
```
ConsoleCatzirl, line 39: Why create the group if the mode is 755? I would expect this to be either 775 or 750 if we are also creating the group.

roles/datadog-builder/tasks/datadog-builder.yml
```diff
@@ -64,3 +78,4 @@
       /opt/venvs/datadog-builder/bin/datadog-builder
       --config /etc/datadog-builder/datadog-secrets.yml
       update {{ datadog_builder_monitor_dir }}/{{ datadog_builder_monitor_file }}
+      2>&1 >> /var/log/datadog-builder/datadog-builder.log
```
ConsoleCatzirl, line 81: Let's also create a logrotate rule for this file.

rattboi:
gate-recheck

PR #117, opened 2017-02-07
Create python-app role

jamielennox:
Being a predominately python shop we install a lot of python
applications from git in very similar ways. We can extract this out into
a role and stop repeating ourselves so much.

### DIFF
--- a/roles/zuul/meta/main.yml
+++ b/roles/zuul/meta/main.yml
@@ -1,2 +1,12 @@
 dependencies:
-  - pip
+  - role: python-app
+    name: zuul
+    python_app_pipdeps:
+      - name: jenkins-job-builder
+        version: '1.6.1'
+      - name: pyzmq
+      - name: ansible
+      - name: statsd
+    python_app_git_repo: "{{ zuul_git_repo_url }}"
+    python_app_git_version: "{{ zuul_git_branch }}"
+    python_app_notify: Restart zuul
--- a/roles/datadog-builder/meta/main.yml
+++ b/roles/datadog-builder/meta/main.yml
@@ -1,0 +1,7 @@
+---
+dependencies:
+  - role: python-app
+    name: datdog-builder
+    python_app_git_repo: "{{ datadog_builder_git_repo_url }}"
+    python_app_git_update: "{{ datadog_builder_git_update }}"
+    python_app_git_version: "{{ datadog_builder_git_branch }}"
--- a/roles/datadog-builder/tasks/datadog-builder.yml
+++ b/roles/datadog-builder/tasks/datadog-builder.yml
@@ -3,22 +3,16 @@
   apt:
     name: "{{ item }}"
     state: installed
-    update_cache: true
   with_items:
-    - git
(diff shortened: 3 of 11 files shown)


roles/zuul/meta/main.yml
```diff
@@ -1,2 +1,12 @@
 dependencies:
-  - pip
+  - role: python-app
+    name: nodepool
```
rattboi: s/nodepool/zuul/

rattboi (dismissed):

roles/zuul/meta/main.yml
rattboi: I confirmed that this is why CI failed. It uses the name for venv, so zuul has /opt/venvs/nodepool, and everything gets confused from there.

rattboi (approved):
:+1: looks good now

roles/zuul/meta/main.yml
```diff
@@ -1,2 +1,12 @@
...
+    name: zuul
+    python_pipdeps:
+      - name: jenkins-job-builder
+        version: '1.6.1'
+      - name: pyzmq
+      - name: ansible
+      - name: statsd
+    git_repo: "{{ zuul_git_repo_url }}"
+    git_version: "{{ zuul_git_branch }}"
+    python_notify: Restart zuul
```
omgjlk: Are you sure this works? You'll be calling the python-app role before the zuul role, and the Restart zuul handler lives in the zuul role. I don't think you'll be able to call it from the python-app role. Please test.

omgjlk (commented):
Overall I think this is okay, but I have a concern about handler calling. Since I'm mostly AFK I won't block this pr with a negative review, but I won't approve it either.

roles/zuul/meta/main.yml
rattboi: https://travis-ci.org/BonnyCI/hoist/jobs/199414443#L1544 Looks like the handler is getting called

jamielennox: So when running it (with -vv) i was seeing output such as: 

```
TASK [python-app : Checkout source] ********************************************
task path: /home/jamie/work/bonny/hoist/roles/python-app/tasks/main.yml:27
NOTIFIED HANDLER Restart zuul
NOTIFIED HANDLER zuul : check zuul-server is running
changed: [jamielennox-playpen] => {"after": "bd8c1adad09f2ea80384d57ac2f873b4ecbf17ba", "before": null, "changed": true, "warnings": []}
```

Which is why i believe this was working, but i had not done an explicit test with a handler that was not being triggered otherwise. I will check and report.

omgjlk: Also make sure that `NOTIFIED HANDLER` is actually followed up with that handler running. Ansible may say that for every thing on a notify, even if there is no matching handler for it. I just don't remember the fine details here.

anne-bonny:
Starting gate_github jobs.

jamielennox:
@j2sol seems like the handlers work fine.

PR #119, opened 2017-02-07
Use a better merger url

ConsoleCatzirl:
Use inventory_hostname instead of ansible_nodename becase the
inventory file uses full FQDNs, but the host doesn't know its own
FQDN.


gandelman-a (approved):

anne-bonny:
Starting gate_github jobs.

PR #121, opened 2017-02-07
Add zuul_connections to all hosts

jamielennox:
If we don't have the github connection in the deploy like in test
scenarios then the github connection isn't available in zuul and the
layout.yaml we add to all hosts isn't there. This means that the
layout.yaml fails validation and zuul doesn't start correctly.

Unfortunately zuul doesn't handle this failure well and instead of
failing completely gets stuck in a half loaded state where nothing
works.


ConsoleCatzirl (commented):
:+1:

anne-bonny:
Starting gate_github jobs.

ConsoleCatzirl (approved):

anne-bonny:
Starting gate_github jobs.

PR #132, opened 2017-02-10
Add apache rewrite rule to also proxy webhooks

gandelman-a:
This updates the apache vhost config on the zuul host to also proxy github
webhooks to the local webapp.  Since our vhost is now doing more than just
serving status data I've updated variable and vhost names.

### DIFF
--- a/install-ci.yml
+++ b/install-ci.yml
@@ -50,8 +50,8 @@
         - monitoring
 
     - role: apache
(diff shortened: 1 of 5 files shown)


ConsoleCatzirl (approved):

ConsoleCatzirl:
gate-recheck

PR #134, opened 2017-02-10
Make it possible to use custom secrets in vagrant

ConsoleCatzirl:
Make the example secrets a fallback for the vagrant environment,
rather than a source of testing truth.

Also, stop trying to install roles from galaxy based on requirements.yml,
that file doesn't exist.


rattboi (approved):
LGTM

PR #135, opened 2017-02-10
Fix empty clouds yaml

rattboi:
This adds the production bastion to the production group, giving it access to the production group_vars, which are needed to populate /etc/openstack/clouds.yaml 

Fixes bonnyci/projman#119


ConsoleCatzirl (approved):
:+1:

PR #157, opened 2017-02-15
Bump our max-servers for contra-sjc

omgjlk:
Our quota was raised, and we need more space

Signed-off-by: Jesse Keating <jkeating@j2solutions.net>


ConsoleCatzirl (approved):
:ok:

PR #158, opened 2017-02-15
Turn on logstash output from filebeat

jamielennox:
There is a default off for outputing filebeat data for logstash meaning
that there is no output configured for the logstash listeners. This
needs to be enabled for the hosts running the filebeat, not the
monitoring hosts which are running logstash itself.

Enable logstash output everywhere as it's setup by default in hoist.


omgjlk (approved):
So there is no danger in having this turned on everywhere for every inventory set?

PR #159, opened 2017-02-16
Fix dd-builder logging in cronjob to use sh syntax

rattboi:
Mistakenly used bash syntax, which doesn't work...

Signed-off-by: Bradon Kanyid <rattboi@gmail.com>


eggshell (approved):

rattboi:
recheck

PR #165, opened 2017-02-21
Add automation to create dev users/projects

omgjlk:
This is a playbook that's meant to be executed by a developer, on a
system that has credentials suitable to create projects and users
defined in a cloud.yaml file. The name of the cloud can be overridden by
setting a runtime variable of "prod_cloud", otherwise it'll default to
"opentech-sjc".

Each user will get a project that matches their username, and get
_member_ role within. Each user will also be added to the admin group,
which defaults to "cloud_admins". This group has cloud_admin role on a
specific project, allowing each member to be able to do cloud_admin
level things should they need to.

This commit also vendors in a newer os_user.py from upstream Ansible, so
that the password will not be updated if the user already exists. This
allows us to set a default password on initial creation, which will
quickly be changed.

Closes BonnyCI/projman#131

Signed-off-by: Jesse Keating <jkeating@j2solutions.net>

### DIFF
--- a/roles/users-projects/tasks/main.yml
+++ b/roles/users-projects/tasks/main.yml
@@ -1,0 +1,39 @@
+---
+- name: ensure projects
+  os_project:
+    cloud: "{{ os_cloud }}"
+    name: "{{ item.username }}"
+  with_items: "{{ users }}"
+
+- name: ensure accounts
+  os_user:
+    cloud: "{{ os_cloud }}"
+    name: "{{ item.username }}"
+    email: "{{ item.email }}"
+    default_project: "{{ item.username }}"
+    password: "{{ default_password }}"
+    update_password: on_create
+  with_items: "{{ users }}"
+
+- name: grant roles
+  os_user_role:
+    cloud: "{{ os_cloud }}"
+    project: "{{ item.username }}"
+    user: "{{ item.username }}"
+    role: "{{ project_role }}"
+  with_items: "{{ users }}"
+
+- name: ensure prod group membership
+  os_user_group:
+    cloud: "{{ os_cloud }}"
+    group: "{{ prod_group }}"
+    user: "{{ item.username }}"
+  with_items: "{{ users }}"
+
+- name: ensure prod admin group membership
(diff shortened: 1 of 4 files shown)


omgjlk:
Don't merge this, I need to adjust it based on the design for production accounts.

roles/users-projects/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    cloud: "{{ os_cloud }}"
+    project: "{{ item.username }}"
+    user: "{{ item.username }}"
+    role: "{{ project_role }}"
+  with_items: "{{ users }}"
+
+- name: ensure prod group membership
+  os_user_group:
+    cloud: "{{ os_cloud }}"
+    group: "{{ prod_group }}"
```
gandelman-a, line 29: i assume this creates the group if it does not already exist?

roles/users-projects/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    user: "{{ item.username }}"
+  with_items: "{{ users }}"
+
+- name: ensure prod admin group membership
+  os_user_group:
+    cloud: "{{ os_cloud }}"
+    group: "{{ prod_admin_group }}"
+    user: "{{ item.username }}"
+  with_items: "{{ users }}"
+  when: "{{ item.isadmin | default('False') | bool }}"
```
gandelman-a, line 39: Do we want to also create the prod projects here or were you saving that for a follow up patch after https://github.com/BonnyCI/bonnyci.org/pull/34 merges?

roles/users-projects/tasks/main.yml
omgjlk: Actually it assumes that the group exists, from the automation that's going to made production projects and groups. A cloud bootstrap type of play.

roles/users-projects/tasks/main.yml
omgjlk: Same answer as above.

gandelman-a (approved):

PR #166, opened 2017-02-21
Add ansible for provisioning network resources on new cloud

gandelman-a:
This adds some new provisioning code for bootstrapping our new cloud
with network resources. It creates a new 'cloud_bootstrap' role that will
supercede the old 'create-hosts' role.  The new one will encompass more than
just hosts and encompass managing production users/projects/roles,  networks,
servers, etc.

Resolves-Issue: BonnyCI/projman#133

### DIFF
--- a/roles/cloud-bootstrap/tasks/main.yml
+++ b/roles/cloud-bootstrap/tasks/main.yml
@@ -1,0 +1,5 @@
+---
+- name: Provision network resources
+  include: network.yml
+  tags:
+    - network
--- a/roles/cloud-bootstrap/tasks/network.yml
+++ b/roles/cloud-bootstrap/tasks/network.yml
@@ -1,0 +1,53 @@
+---
+- name: Create control-plane network
+  os_network:
+    cloud: "{{ cloud }}"
+    name: control-plane
+
+- name: Create control-plane subnet
+  os_subnet:
+    cloud: "{{ cloud }}"
+    name: control-plane-subnet
+    network_name: control-plane
+    cidr: "{{ controlplane_network_cidr }}"
+
+- name: Create nodepool network
+  os_network:
+    cloud: "{{ cloud }}"
+    name: nodepool
+    shared: true
+
+- name: Create nodepool subnet
+  os_subnet:
+    cloud: "{{ cloud }}"
+    name: nodepool-subnet
+    network_name: nodepool
+    cidr: "{{ nodepool_network_cidr }}"
+
+- name: Create router
+  os_router:
+    cloud: "{{ cloud }}"
+    name: bonnyci-router
+    network: "{{ external_network_name }}"
+    interfaces:
+        - control-plane-subnet
+        - nodepool-subnet
+
+- name: Create security groups
+  os_security_group:
+    cloud: "{{ cloud }}"
+    name: "{{ item.name }}"
+  with_items: "{{ security_group_rules }}"
+
+- name: Create security group rules
+  os_security_group_rule:
+    cloud: "{{ cloud }}"
+    security_group: "{{ item.0.name }}"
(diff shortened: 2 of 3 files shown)


roles/cloud-bootstrap/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: Provision network resources
+  include: network.yml
+  tags:
+    - network
```
eggshell, line 5: Is there a plan to add any other sections to this? If the `cloud-bootstrap` role is just meant for networking, then it would be good to stay with convention and keep all tasks in `main.yml`. If we're planning on having things like user and tenant creation here eventually, then it's fine as is.

roles/cloud-bootstrap/tasks/network.yml
```diff
@@ -0,0 +1,66 @@
+---
+- name: Create control-plane network
+  os_network:
+    cloud: "{{ cloud }}"
```
eggshell, line 4: Do we have a default definition for `cloud`? I see `os_cloud` in the `create_hosts` role that this replaces, and a `cloud` attribute of `nodepool_providers`.

eggshell (commented):
This looks pretty sane to me and matches up with the design doc. I have a few questions posted inline.

roles/cloud-bootstrap/tasks/network.yml
```diff
@@ -0,0 +1,66 @@
...
+    security_group: "{{ item.0.name }}"
+    protocol:  "{{ item.1.protocol | default('tcp') }}"
+    port_range_min: "{{ item.1.port_min | default(item.1.port)  }}"
+    port_range_max: "{{ item.1.port_max | default(item.1.port) }}"
+    remote_group: "{{ item.1.remote_group }}"
+  with_subelements:
+    - "{{ security_group_rules }}"
+    -  "rules"
+  when:  item.1.remote_group is defined
```
omgjlk: These two tasks can be combined into one, making use of the `omit` filter.

```
  remote_group: "{{ item.1.remote_group | default('omit') }}"
  remote_ip_prefix: "{{ item.1.remote_cidr | default('omit') }}"
```

This way depending on the rule, either `remote_group` or `remote_ip_prefix` will be used.

roles/cloud-bootstrap/tasks/main.yml
omgjlk: I think, if I read the message right, this is the start of the role, and more tasks will be added to it in future PRs. This particular one just tackles the network aspect of it.

omgjlk (changes requested):
I'm a bit concerned about the security groups. If you create the security group that the node pool nodes need, will the node pool project be able to see that group? It can see the network, as the network is shared, but what about the security group?

roles/cloud-bootstrap/tasks/main.yml
eggshell: yep. disregard my comment above.

gandelman-a:
@j2sol nodepool doenst need to see any of these security groups, it will only ever be booting instances on the nodepool network with the default security group. The security group relevant to the slaves here is the sg-zuul-merger, which allows inbound git clones to the merges from the nodepool network, but again the nodepool side doesnt need to be concerned with that.

omgjlk:
@gandelman-a that's an excellent point! I rescind my concern.

I'll wait for the spec to get approved and then lift my negative review.

roles/cloud-bootstrap/tasks/network.yml
gandelman-a: Cool tip, updated.

omgjlk (approved):

PR #167, opened 2017-02-22
Add mctaylor to add-users-projects.yml

eggshell:
I want to be able to have a vm on the cloud, and shouldn't need admin
rights to do that.

### DIFF
--- a/create-users-projects.yml
+++ b/create-users-projects.yml
@@ -9,3 +9,6 @@
         - username: jkeating
           email: "{{ username + '@us.ibm.com' }}"
           isadmin: True
+        - username: mctaylor
+          email: "{{ username + '@us.ibm.com' }}"
+          isadmin: False


create-users-projects.yml
```diff
@@ -9,3 +9,6 @@
         - username: jkeating
           email: "{{ username + '@us.ibm.com' }}"
           isadmin: True
+        - username: mctaylor
+          email: "{{ mctaylor + '@us.ibm.com' }}"
```
omgjlk: `mctaylor` is assumed to be a variable there, not a literal string. I think you want to either use `username` as the variable, or quote `'mctaylor'`.

omgjlk (changes requested):
The code needs changes, but also blocking until we have the base project/groups in place that you would be added to.

create-users-projects.yml
eggshell: whoops, thanks

omgjlk (approved):
Going to let this in, nothing will execute it until we're ready anyway.

PR #168, opened 2017-02-23
Add ansible for provisioning projects,users,groups

eggshell:
This adds the ability to provision our desired projects, groups and
users on the new cloud. It is a continuation of the cloud_bootstrap
role started in BonnyCI/hoist#166, and follows the spec contained in
BonnyCI/bonnyci.org#34.

Depends-On: BonnyCI/hoist#166
Resolves: BonnyCI/projman#134


create-users-projects.yml
```diff
@@ -2,9 +2,9 @@
 - name: Create users and projects
   hosts: localhost
   roles:
-    - role: users-projects
-      tags: ['create-hosts']
-      os_cloud: opentech-sjc
+    - role: cloud-bootstrap
+      tags: ['create-users-projects']
```
omgjlk: If we're going to make this more of a bootstrap role, we should remove the users-projects tag off of the role itself, and instead rely on the tags inside the role.

Should probably re-name the play file too.

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: Create bonnyci project
+  os_project:
+    cloud: "{{ cloud }}"
+    name: bonnyci
```
omgjlk: This should be a variable from the role defaults.

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- name: Create bonnyci group
+  os_group:
+    cloud: "{{ cloud }}"
+    name: bonnyci
```
omgjlk: Same with this, a role variable.

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- name: Create bonnyci-admin group
+  os_group:
+    cloud: "{{ cloud }}"
+    name: bonnyci-admin
```
omgjlk: Same.

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- name: Add all users to bonnyci group
+  os_user_group:
+    cloud: "{{ cloud }}"
+    group: bonnyci
```
omgjlk: variable

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: Create bonnyci project
```
omgjlk: the task names should be more generic, such as the `production project` or some such. Other implementers of this stuff may not want to call it bonnyci.

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+  os_user_group:
+    cloud: "{{ cloud }}"
+    group: bonnyci
+    user: "{{ item.username }}"
+  with_items: users
+
+- name: Add all bonnyci group members to bonnyci project
+  os_user_role:
+    cloud: "{{ cloud }}"
+    role: member
```
omgjlk: this role should be `_member_`, but also should be a variable.

In fact, the existing role `users-projects` did this bit. Your commit sort of leaves that role an orphan. Could that code instead be used in this role?

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+    group: bonnyci
+    project: bonnyci
```
omgjlk: These should be variables.

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- name: Add admin users to bonnyci-admin group
+  os_user_group:
+    cloud: "{{ cloud }}"
+    group: bonnyci-admin
```
omgjlk: variable

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+    user: "{{ item.username }}"
+  with_items: users
+  when: item.isadmin == True
```
omgjlk: It's preferable to use the `bool` filter to check for truthiness. Also, since this attribute might only be set WHEN a user is an admin, we have to handle the default case where it is not defined:

`when: item.isadmin | default('False') | bool`

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- name: Grant bonnyci-admin group cloud_admin on bonnyci project
+  os_user_role:
+    cloud: "{{ cloud }}"
+    role: cloud_admin
+    group: bonnyci-admin
+    project: bonnyci
```
omgjlk: variables.

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+    name: "{{ item.username }}"
+  with_items: users
+
+- name: Assign users to their user project
+  os_user_role:
+    cloud: "{{ cloud }}"
+    role: member
+    user: "{{ item.username }}"
+    project: "{{ item.username }}"
+  with_items: users
```
omgjlk: all of these need variables.

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+  os_group:
+    cloud: "{{ cloud }}"
+    name: bonnyci-admin
+
+- name: Add all users to bonnyci group
+  os_user_group:
+    cloud: "{{ cloud }}"
+    group: bonnyci
+    user: "{{ item.username }}"
+  with_items: users
```
omgjlk: With ansible 2.x, the argument to loops needs to be templated.

`with_items: {{ users }}`

roles/cloud-bootstrap/tasks/main.yml
```diff
@@ -3,3 +3,8 @@
   include: network.yml
   tags:
     - network
+
+- name: Provision users and projects
+  include: create-users-projects.yml
+  tags:
+    - create-users-projects
```
omgjlk: since the previous tag in this role is more simplistic, `network`, we could be simple in this tag as well, `accounts`. or some such.

However, since the network bits require the users/projects in place, this new task set to create the users and projects needs to run before the network tasks.

omgjlk (changes requested):
Comments in-line.

roles/cloud-bootstrap/tasks/main.yml
gandelman-a: So actually this is something we'll need to figure out once the network stuff merges... is how we create the bonnyci project here, and use it in network.yml when creating those resources. Currently sifting thru the os modules+shade+os-client-config to see if there's a way we can just specify it in the play, but it looks like we may need to keep multiple clouds.yaml entries for each project

gandelman-a: Actually my point looks easily resolved. Once we create the project, we can pass an auth parameter to the resources in network.yml, that specifies only the project and it will override in the auth provided the user running the play has access.

anne-bonny:
Build failed.

- http://logs.bonnyci.com/check_github/BonnyCI/hoist/168/1487885728.63/bonnyci-run-check-multinode : FAILURE in 56s

roles/cloud-bootstrap/tasks/network.yml
```diff
@@ -0,0 +1,66 @@
+---
+- name: Create control-plane network
+  os_network:
+    cloud: "{{ cloud }}"
+    name: control-plane
```
gandelman-a: So in all of these os_* tasks, we need to specify the project we want to create them in with an auth: parameter, ie:
```
auth:
  project_name: bonnyci
```

eggshell: @gandelman-a cool. I'll switch up the order of `accounts.yml` and `network.yml` and add the auth parameters.

roles/cloud-bootstrap/tasks/create-users-projects.yml
```diff
@@ -0,0 +1,83 @@
...
+- name: Create nodepool project
+  os_project:
+    cloud: "{{ cloud }}"
+    name: nodepool
+
+- name: Create nodepool user
+  os_user:
+    cloud: "{{ cloud }}"
+    name: nodepool
+    default_project: nodepool
```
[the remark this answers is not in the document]
eggshell: @gandelman-a I used `secrets.nodepool.db_password` since I didn't see just a `password` var in the example secrets.

roles/cloud-bootstrap/tasks/create-users-projects.yml
eggshell: @j2sol I think the `user-projects` role assumes a slightly different approach to our account provisioning, but I do agree that the code is extremely similar. Do you think we should remove that once this is merged and fully tested?

omgjlk (changes requested):
getting better, still some problems.

create-users-projects.yml
```diff
@@ -2,9 +2,9 @@
 - name: Create users and projects
   hosts: localhost
   roles:
-    - role: users-projects
-      tags: ['create-hosts']
-      os_cloud: opentech-sjc
+    - role: cloud-bootstrap
+      tags: ['accounts']
```
omgjlk: This tag is inappropriate, as it applies the `accounts` tag to the entire role, and inside that role, both accounts _and_ networks are created. It'd be best to remove the tag here all together, and just rely on the tags inside the `main.yml` file within the role.

roles/cloud-bootstrap/tasks/accounts.yml
```diff
@@ -0,0 +1,76 @@
+---
+- name: Ensure projects
+  os_project:
+    cloud: "{{ cloud }}"
+    name: "{{ item }}"
+  with_items:
+    - "{{ prod_project }}"
+    - "{{ nodepool_project }}"
+    - "{{ users.username }}"
```
omgjlk: This will probably fail. `users.username` has no meaning at this point. I think you may want to do some concatenation for the three variable sets, or since that's awkward, just do one task that has the `prod_project` and `nodepool_project` items, and a second task that loops over `users` and uses `item.username`. No need to be really clever, just make it clear to understand.

omgjlk (changes requested):
really close now! Just need to fix up tags and some of the loops, and then I think you can just delete all of the `users-projects` role as part of this change.

roles/cloud-bootstrap/tasks/network.yml
```diff
@@ -3,26 +3,34 @@
   os_network:
     cloud: "{{ cloud }}"
     name: control-plane
+    auth:
+      project_name: "{{ prod_project }}"
```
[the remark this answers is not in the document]
gandelman-a: Right but this will be run from an admin locally and may have a different project (their personal) set on the cloud their using.

gandelman-a:
I think we still need the ``users-projects`` role, to manage the personal dev and admin accounts that will eventually be running this?  We could potentially role it into the ``cloud-bootstrap`` role/

gandelman-a:
This LGTM, will let @j2sol look and change his negative to a positive.

ConsoleCatzirl (commented):
:star: looks good to me, but someone else should approve

roles/cloud-bootstrap/tasks/accounts.yml
```diff
@@ -0,0 +1,81 @@
...
+  with_items:
+    - "{{ prod_group }}"
+    - "{{ prod_admin_group }}"
+
+- name: Ensure users
+  os_user:
+    cloud: "{{ cloud }}"
+    name: "{{ item.username }}"
+    email: "{{ item.email }}"
+    default_project: "{{ item.username }}"
```
omgjlk: This needs to set a password, defaulting to a simple one, like in the previous role. It also needs to set the task to not update the password on subsequent runs.

```
     password: "{{ default_password }}"
     update_password: on_create
```

roles/cloud-bootstrap/tasks/accounts.yml
```diff
@@ -0,0 +1,81 @@
...
+  with_items: "{{ users }}"
+
```
omgjlk: Should copy over the task that grants the user a role on their project, otherwise they won't be able to do anything.

Grab the task from https://github.com/BonnyCI/hoist/pull/168/files#diff-e0b563178351dc3acf8378b82947655eL18 and make sure any defaults used are ported over as well.

omgjlk (changes requested):

mlangbehn (commented):
Looks good to me. @j2sol, did you get a chance to look back over this yet?

omgjlk (approved):
Woo thanks for sticking it out on this one, lots of back and forth. Lets see how it blows up!

PR #171, opened 2017-03-01
Use python-app variables for os_loganalyze venv path

jamielennox:
The python-app role will set the variables for the venv and source dir
so that dependant roles can use them. This means you aren't going to end
up with a mispelt path - like we have now.


ConsoleCatzirl (commented):
LGTM, but not approving until after the CNCF demo.

ConsoleCatzirl (approved):

PR #173, opened 2017-03-01
Move apache status page to a different port

jamielennox:
The apache status module helpfully installs itself by default onto all
vhosts, but restricts itself to localhost. This is questionable
behaviour anyway but it does not play nicely with our log rendering on
the logs server and so dd-apache is failing on that machine.

Construct a new vhost on a random port that only listens on localhost
and serves the status url. Then point the datadog check to this port
instead.

### DIFF
--- a/roles/apache/templates/vhost.conf
+++ b/roles/apache/templates/vhost.conf
@@ -2,7 +2,11 @@
 
 {{ item.global_vhost_settings | default("") }}
 
+{% if item.listen_host | default(False) %}
+Listen {{ item.listen_host }}:{{ item.port | default(default_port) }}
+{% else %}
 Listen {{ item.port | default(default_port) }}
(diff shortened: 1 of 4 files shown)


ConsoleCatzirl (commented):
:star: looks good to me

eggshell (commented):
lgtm

roles/apache/templates/vhost.conf
```diff
@@ -2,7 +2,11 @@
 
 {{ item.global_vhost_settings | default("") }}
 
+{% if item.listen_host | default(False) %}
+Listen {{ item.listen_host  }}:{{ item.port | default(default_port) }}
```
eggshell: extra space in `{{ item.listen_host }}`

omgjlk (approved):

PR #181, opened 2017-03-02
Tighten up network resource creation

gandelman-a:
The openstack modules and shade resolve things via name by default.
This is a problem if running automation with admin credentials and
other tenants have created resources with the same name, as admin
can view all of them and automation will blow up on colliding names.

This tightens it up by registering ids and passing those instead of
names. It also passes the 'project' argument to further
tighten its queries.

Also, this creates the required hole in nodepool's default security
group so things can reach its slaves via ssh from the control-plane.


ConsoleCatzirl (commented):
:+1:

eggshell (approved):

PR #182, opened 2017-03-02
Fix ansible-runner role showing changed every run

mlangbehn:
Previously, the ansible-runner role had four tasks that reported as
changed on subsequent runs (three of them reported as changed twice per
run). Now the tasks have been updated so that they do not report as
changed on subsequent runs.

Related-Issue: BonnyCI/projman#38

### DIFF
--- a/roles/ansible-runner/tasks/main.yml
+++ b/roles/ansible-runner/tasks/main.yml
@@ -27,20 +27,27 @@
   file:
     path: "{{ ansible_runner_virtualenv }}/plugins/callback"
     state: directory
-    owner: "{{ ansible_runner_user }}"
+    mode: 0755
+    owner: root
+    group: root
 
 - block:
   - name: Install datadog callback plugin
     get_url:
       url: "{{ datadog_callback_url }}"
       dest: "{{ ansible_runner_virtualenv }}/plugins/callback/datadog_callback.py"
-      owner: "{{ ansible_runner_user }}"
+      mode: 0755
+      owner: root
+      group: root
 
   - name: Install datadog callback configuration
     copy:
       dest: "{{ ansible_runner_virtualenv }}/plugins/callback/datadog_callback.yml"
       content: "api_key: {{ secrets.datadog.api_key }}"
-      owner: "{{ ansible_runner_user }}"
+      mode: 0644
+      owner: root
+      group: root
+
   when: datadog_enabled | default(True) | bool
   tags:
     - datadog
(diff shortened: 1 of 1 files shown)


roles/ansible-runner/tasks/main.yml
```diff
@@ -74,13 +81,6 @@
     dest: "/opt/source/{{ ansible_runner_job }}"
     update: true
 
-- name: Set source directory owner
-  file:
-    state: directory
-    path: "/opt/source/{{ ansible_runner_job }}"
-    recurse: yes
-    owner: "{{ ansible_runner_user }}"
-
```
rattboi, line 83: This should be fine to remove because all the files are +r for other, and cideploy can read them.

roles/ansible-runner/tasks/main.yml
```diff
@@ -27,20 +27,27 @@
...
+      mode: 0755
+      owner: root
+      group: root
 
   - name: Install datadog callback configuration
     copy:
       dest: "{{ ansible_runner_virtualenv }}/plugins/callback/datadog_callback.yml"
       content: "api_key: {{ secrets.datadog.api_key }}"
-      owner: "{{ ansible_runner_user }}"
+      mode: 0755
```
rattboi: This should be 0644, no?

rattboi (changes requested):
Shouldn't set the datadog yml to be executable when it isn't executable.

rattboi:
Also, this doesn't look like it fixes the fact that secrets permissions are changed every run, right? Since this role is run twice, the owner flipflops every run, and the last one is the winner...

roles/ansible-runner/tasks/main.yml
ConsoleCatzirl: Ya, this should be 0644

ConsoleCatzirl (commented):
Since this changes the owner from ansible_runner_user to root, I don't think the owner will flip-flop any more.

roles/ansible-runner/tasks/main.yml
mlangbehn: @rattboi: Thanks. I made a mental note to set it to 0644, but never actually made the change.

mlangbehn:
@rattboi: No, there is a [separate task](https://github.com/BonnyCI/hoist/blob/master/roles/ansible-runner/tasks/secret-file.yml#L6) that sets the secrets file to `root:root 0400`. There is [another task](https://github.com/BonnyCI/hoist/blob/master/roles/ansible-runner/tasks/secret-file.yml#L28) which makes sure that the `"{{ ansible_runner_user }}"` has a read allow ACE set on the secrets file, but that will not cause any flip-flopping of permissions (since it is just an entry in the ACL).

**EDIT:**
I started digging into the permissions changes on the secrets file. The issue is outside the scope of this pull request, but I have opened BonnyCI/projman#154 to track it.

eggshell (commented):
looks good to me. want to let @rattboi and @jesusaurus have another look though.

ConsoleCatzirl (commented):
:+1:

rattboi (approved):

PR #183, opened 2017-03-03
Create a zuul user on the logs server

jamielennox:
The zuul launcher will try to rsync from the worker to the logs server
as the zuul user. Somehow this is present in production but not created
by hoist.

Add a task that simply creates a user and adds the authorized key so
that logs can be uploaded.

### DIFF
--- a/install-ci.yml
+++ b/install-ci.yml
@@ -15,7 +15,7 @@
         - letsencrypt
       when:
         - letsencrypt_csr_cn | default(False)
-        
+
 - name: Install ELK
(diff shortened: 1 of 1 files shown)


jamielennox:
I'm sure there are nicer ways to do this, but it doesn't seem worth a role and I'm not sure what we can rely on for secrets at the moment. 

For now i just want to put this here as a reminder how i got past the problem and if someone wants to make it prettier you should feel free.

ConsoleCatzirl (approved):

PR #184, opened 2017-03-03
Create an ansible role to install quartermaster

rattboi:
Installs quartermaster, our standup bot

This role makes a few assumptions.

1) This isn't managed with ansible-runner like the rest of our stack,
as it shouldn't be part of the critical path for our production installs
2) This is meant to be run from the bastion directly
3) secrets need to be in-place beforehand

secrets:
  quartermaster:
    defaults:
      <users>
      <groups>
      <timezones>

Fixes BonnyCI/projman#153

### DIFF
--- a/roles/quartermaster/tasks/main.yml
+++ b/roles/quartermaster/tasks/main.yml
@@ -1,0 +1,85 @@
+---
+- name: install dependencies
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - unzip
+
+- name: Create the quartermaster group
+  group:
+    name: quartermaster
+    state: present
+
+- name: Create quartermaster user
+  user:
+    name: quartermaster
+    group: quartermaster
+    home: /var/lib/quartermaster
+    system: yes
+    state: present
+
+- name: Create directories
+  file:
+    path: "{{ item }}"
+    state: directory
+    mode: 0755
+    owner: quartermaster
+    group: quartermaster
+  with_items:
+    - /etc/quartermaster
+    - /var/log/quartermaster
+    - /opt/quartermaster
+
+- name: Download quartermaster binary
+  get_url:
+    url: "{{ quartermaster_install_url }}"
+    dest: /tmp
+
+- name: Install quartermaster binary
+  unarchive:
+    src: "/tmp/{{ quartermaster_install_zip_filename }}"
+    dest: /opt/quartermaster
+    remote_src: True
+
+- name: Install quartermaster config
+  template:
+    src: etc/quartermaster/config.yaml
+    dest: /etc/quartermaster/config.yaml
+    owner: quartermaster
+    mode: 0600
+  notify: Restart quartermaster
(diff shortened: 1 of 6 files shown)


rattboi:
If someone wants to test it, https://github.com/rattboi/quartermaster-role/ has the role on its own, as well as instructions on using docker/docker-compose to run it.

roles/quartermaster/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
...
+    owner: root
+    group: root
+    mode: 0644
+  with_items:
+    - quartermaster
+  notify:
+    - Reload systemd units
+    - Restart quartermaster
+
+- meta: flush_handlers
```
eggshell, line 79: Since `Restart quartermaster` has been notified up to this point and flush_handlers is being used, the quartermaster service will always be bounced at this point. So the next task in a fresh install will only enable it. Not really a big deal, but wanted to bring it up.

roles/quartermaster/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
...
+
+- name: Enable and start systemd service
+  service:
+    enabled: yes
+    state: running
```
eggshell: `state: started`

eggshell (commented):
Looking great! Left a few comments inline.

roles/quartermaster/tasks/main.yml
pschwartz: This appears to be the only change needed.

pschwartz (dismissed):

roles/quartermaster/tasks/main.yml
rattboi: Interesting. This works in my tests. Maybe it's just not as accurate?

roles/quartermaster/tasks/main.yml
rattboi: This doesn't seem to restart on every run in my test environment. On the 2nd run, if the previous tasks didn't change (no systemd unit file changes, no new config), then this isn't triggered.

roles/quartermaster/tasks/main.yml
rattboi: https://github.com/BonnyCI/hoist/blob/master/roles/nodepool/tasks/main.yml#L161 is one of our files I referenced to create this role...

anne-bonny:
Build failed.

- http://logs.bonnyci.com/check_github/BonnyCI/hoist/184/1488563521.18/bonnyci-run-check-multinode : FAILURE in 2m 36s

pschwartz (approved):

rattboi:
recheck

anne-bonny:
Build failed.

- http://logs.bonnyci.com/check_github/BonnyCI/hoist/184/1488564075.04/bonnyci-run-check-multinode : FAILURE in 48s

rattboi:
recheck

PR #190, opened 2017-03-06
Add ubuntu-user role

SpamapS:


anne-bonny:
Build failed.

- http://logs.bonnyci.com/check_github/BonnyCI/hoist/190/1488825967.94/bonnyci-run-check-multinode : FAILURE in 1m 00s

roles/common/meta/main.yml
```diff
@@ -1,4 +1,5 @@
 dependencies:
+  - admin-user
```
ConsoleCatzirl: The common role is also run on the bastion host, and I would prefer to not include this user on that host. Can we move this role to install-ci.yml instead of being a dependency of the common role?

ConsoleCatzirl (changes requested):

roles/common/meta/main.yml
SpamapS: Sounds like a good idea. I'll move it.

ConsoleCatzirl:
:+1: looks good to me

eggshell (commented):
looks good to me.

ConsoleCatzirl (approved):

PR #193, opened 2017-03-07
Prepare to use the new cloud

ConsoleCatzirl:
First, create a script in the tools/ directory for installing ansible dependencies so that we don't need to copy-paste the command when bootstrapping new bastions.

Second, refactor most of our production bits to support multiple parallel production deployments.


omgjlk (approved):
Looks reasonable to me.

ConsoleCatzirl:
recheck

PR #195, opened 2017-03-07
Move old letsencrypt cert when requesting a new one

ConsoleCatzirl:
The letsencrypt module will check the validity of the current cert before
requesting a new one, without checking that the cert and key match. If a
new key and cert request are generated while the old cert is still valid,
this leaves us with a key/cert mismatch and apache will fail to start.

If we generate a new key and/or cert request, move the old cert to a backup
location to force the letsencrypt module to generate a new cert.

### DIFF
--- a/roles/letsencrypt/tasks/main.yml
+++ b/roles/letsencrypt/tasks/main.yml
@@ -50,6 +50,19 @@
         (not letsencrypt_stat.results.1.stat.exists) or
         (letsencrypt_create_csr.changed | default(False))
 
+- name: Backup old cert when requesting a new one
+  copy:
+    remote_src: True
+    src: "{{ letsencrypt_cert_path }}"
+    dest: "{{ letsencrypt_cert_path }}.bak"
+  when: letsencrypt_create_csr_command.changed | default(False)
+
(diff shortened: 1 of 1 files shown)


roles/letsencrypt/tasks/main.yml
```diff
@@ -50,6 +50,19 @@
         (not letsencrypt_stat.results.1.stat.exists) or
         (letsencrypt_create_csr.changed | default(False))
 
+- name: Backup old cert when requesting a new one
+  file:
+    remote_src: True
+    src: "{{ letsencrypt_cert_path }}"
+    dest: "{{ letsencrypt_cert_path }}.bak"
+  when: letsencrypt_create_csr_command.changed
```
eggshell: Should this have a default case?

eggshell (commented):
Overall looks good, have one question inline.

roles/letsencrypt/tasks/main.yml
ConsoleCatzirl: Yes, it should. Good catch :)

ConsoleCatzirl:
recheck

eggshell (approved):

PR #200, opened 2017-03-08
Clean up old provisioning roles, old ssh users

gandelman-a:
This cleans up some old provisioning code that is superseded
by the cloud-bootstrap role. It also cleans out some old users
from the bastion ssh users.


roles/bastion/defaults/main.yml
```diff
@@ -16,5 +16,6 @@ bastion_users:
   - jesusaur
   - eggshell
 bastion_sudo_users: "{{ bastion_users }}"
-bastion_remove_users: []
+bastion_remove_users:
+  - pschwartz
```
rattboi, line 19: You need to remove the user from bastion_users as well

rattboi (changes requested):

rattboi (approved):

PR #201, opened 2017-03-08
Add datadog check for stale nodepool images

eggshell:
Feeds age of all images in nodepool dib-image-list.json to datadog.

Implements: BonnyCI/projman#98

### DIFF
--- a/roles/dd-nodepool/defaults/main.yml
+++ b/roles/dd-nodepool/defaults/main.yml
@@ -1,2 +1,3 @@
 ---
 dd_nodepool_status_url:  http://locahost:8005/image-list
+dd_nodepool_images_url:  http://locahost:8005/dib-image-list.json
--- a/roles/dd-nodepool/files/nodepool_server.py
+++ b/roles/dd-nodepool/files/nodepool_server.py
@@ -20,24 +20,40 @@ from checks import AgentCheck
 class NodepoolCheck(AgentCheck):
 
     SERVICE_CHECK_NAME = 'nodepool_server.got_status'
+    IMAGES_CHECK_NAME = 'nodepool_server.got_images'
 
     def check(self, instance):
-        if 'status_url' not in instance:
-            self.log.info("Skipping instance, no url found.")
-            return
-
-        status_url = instance['status_url']
         default_timeout = self.init_config.get('default_timeout', 5)
         timeout = float(instance.get('timeout', default_timeout))
         service_check_tags = ['url:%s' % status_url]
-        
-        try:
-            self.get_nodepool_status(status_url, timeout, service_check_tags)
-        except Exception as e:
-            self.service_check(self.SERVICE_CHECK_NAME,
-                               AgentCheck.CRITICAL,
-                               message=str(e),
-                               tags=service_check_tags)
+        images_check_tags = ['url:%s' % images_url]
+
+        status_url = instance.get('status_url')
+        image_url = instance.get('image_url')
+
+        if status_url:
+            try:
+                self.get_nodepool_status(status_url, timeout,
+                                         service_check_tags)
+            except Exception as e:
+                self.service_check(self.SERVICE_CHECK_NAME,
+                                   AgentCheck.CRITICAL,
+                                   message=str(e),
+                                   tags=service_check_tags)
+        else
+            self.log.info("Skipping instance, no status url found.")
+
+        if images_url:
+            try:
+                self.get_nodepool_images(images_url, timeout,
+                                         images_check_tags)
+            except Exception as e:
+                self.service_check(self.IMAGES_CHECK_NAME,
+                                   AgentCheck.CRITICAL,
+                                   message=str(e),
+                                   tags=images_check_tags)
+        else
+            self.log.info("Skipping instance, no images url found.")
 
     def get_nodepool_status(self, status_url, timeout, service_check_tags):
         r = requests.get(status_url, timeout=timeout)
@@ -55,3 +71,10 @@ class NodepoolCheck(AgentCheck):
                            AgentCheck.OK,
                            message='Received status from %s' % status_url,
                            tags=service_check_tags)
+
+    def get_nodepool_images(self, images_url, timeout, images_check_tags):
+        r = requests.get(images_url, timeout=timeout)
+        currentTime = int(time.time())
+
+        for image in r.json():
+            self.gauge('nodepool.image.%s.age' % image['name'], currentTime - image['age'])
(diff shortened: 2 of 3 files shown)


roles/dd-nodepool/defaults/main.yml
```diff
@@ -1,2 +1,3 @@
 ---
 dd_nodepool_status_url:  http://locahost:8005/image-list
+dd_nodepool_images_url:  http://locahost:8005/dib-image-list
```
gandelman-a: This url gives a human readable table with the images in it.  It should hit /dib-image-list.json and that will give you something you can easily look at below.

roles/dd-nodepool/files/nodepool_server.py
```diff
@@ -55,3 +70,25 @@ def get_nodepool_status(self, status_url, timeout, service_check_tags):
                            AgentCheck.OK,
                            message='Received status from %s' % status_url,
                            tags=service_check_tags)
+
+    def get_nodepool_images(self, images_url, timeout, service_check_tags):
+        r = requests.get(images_url, timeout=timeout)
+        currentTime = int(time.time())
+        staleImage = False
+
+        for image in r:
```
gandelman-a: You're iterating through the lines of the body here.  If you update the endpoint and request the json, you can instead do:

```
for image in r.json():
```

... which'll iterate instead through image dicts below.

roles/dd-nodepool/files/nodepool_server.py
```diff
@@ -55,3 +70,25 @@ def get_nodepool_status(self, status_url, timeout, service_check_tags):
...
+            if currentTime - image.age > 172800:
```
gandelman-a: In either case, this'll be an attribute error.  If updated to do the json thing, you can access dict items instead:

``
if currentTime - image['age'] > 172800:
``

roles/dd-nodepool/files/nodepool_server.py
```diff
@@ -55,3 +70,25 @@ def get_nodepool_status(self, status_url, timeout, service_check_tags):
...
+                self.service_check(
+                    self.IMAGES_CHECK_NAME,
+                    AgentCheck.CRITICAL,
+                    message='Nodepool image is stale: %s' % image.id
```
gandelman-a: same as above

gandelman-a (dismissed):
Cool, we need this.  A few comments and fixes requested.

roles/dd-nodepool/files/nodepool_server.py
```diff
@@ -20,17 +20,24 @@
...
 
     def check(self, instance):
         if 'status_url' not in instance:
-            self.log.info("Skipping instance, no url found.")
+            self.log.info("Skipping instance, no status url found.")
+            return
+
+        if 'images_url' not in instance:
+            self.log.info("Skipping instance, no images url found.")
             return
```
jamielennox: we probably don't want to return from both of these. If status_url isn't found then skip the status part, if images_url isn't found skip the images part.

roles/dd-nodepool/files/nodepool_server.py
```diff
@@ -55,3 +70,25 @@ def get_nodepool_status(self, status_url, timeout, service_check_tags):
...
+                    self.IMAGES_CHECK_NAME,
+                    AgentCheck.CRITICAL,
+                    message='Nodepool image is stale: %s' % image['id']
+                    tags=images_check_tags)
+                staleImage = True
+
+        if staleImage:
+            return
+
+        self.service_check(self.IMAGES_CHECK_NAME,
```
jamielennox: I don't think you want this to be a service check. A service check is a good/bad check for when something has actually gone wrong. What you can do here is report to datadog the actual age of the image and then on the datadog side we say that it is something we need to be warned about if age > X. I think you probably want a gauge: http://docs.datadoghq.com/guides/dogstatsd/#metrics

jamielennox (changes requested):

roles/dd-nodepool/files/nodepool_server.py
```diff
@@ -20,24 +20,39 @@
...
     SERVICE_CHECK_NAME = 'nodepool_server.got_status'
+    IMAGES_CHECK_NAME = 'nodepool_server.got_images'
 
     def check(self, instance):
-        if 'status_url' not in instance:
-            self.log.info("Skipping instance, no url found.")
-            return
-
         status_url = instance['status_url']
+        images_url = instance['images_url']
```
jamielennox: you removed the if status_url in instance check, so these will KeyError if the value isn't present in config. Just move it down and use instance['status_url'] instead of status_url in the if block.

roles/dd-nodepool/files/nodepool_server.py
```diff
@@ -55,3 +70,10 @@ def get_nodepool_status(self, status_url, timeout, service_check_tags):
                            AgentCheck.OK,
                            message='Received status from %s' % status_url,
                            tags=service_check_tags)
+
+    def get_nodepool_images(self, images_url, timeout, images_check_tags):
+        r = requests.get(images_url, timeout=timeout)
+        currentTime = int(time.time())
+
+        for image in r.json():
+            self.gauge(self.IMAGES_CHECK_NAME, currentTime - image['age'])
```
jamielennox: hmm, so i'm not entirely sure what to do here, but this won't work. If you submit each metric with the same name we will end up in datadog with a bunch of values under 'nodepool.got_images' that you can't correlate to each other.  I don't think anyone here is particularly familiar with how we want this reported in datadog yet, but we at least need to identify the image name some how. So at a first guess: 

`self.gauge('nodepool.image.%s.age' % image['name'], currentTime - image['age'])`

Also without looking I'm surprised that the entry 'age' is an absolute time, rather than an age already.

jamielennox (changes requested):

roles/dd-nodepool/files/nodepool_server.py
eggshell: @jamielennox when I first started taking a look at this, the image age being an absolute time was the first point of contention. If you hop on prod and look at the age of an image, it should be an absolute time, which is why this check has to compare that against the current time to get an accurate value.

jamielennox (changes requested):

anne-bonny:
Build failed.

- http://logs.bonnyci.org/check_github/BonnyCI/hoist/201/1489612188.4/bonnyci-run-check-multinode : FAILURE in 7m 46s

eggshell:
recheck

jamielennox (approved):

jamielennox:
Thanks for sticking with it.

PR #212, opened 2017-03-10
Don't add integration key file to config if no data

jamielennox:
If there is no integration key contents defined in secrets we shouldn't
add a non-existant file to the zuul config. This is causing an error in
the github connection that is crashing the main thread of zuul.


jamielennox:
You know you have decent CI when you upload pull requests and let your CI test if it works, rather than do it yourself.

rattboi (approved):

PR #217, opened 2017-03-12
Add a requests logger to zuul-server

jamielennox:
Split up the different zuul logging files and add a new logger to the zuul
server that logs the requests that are made.

### DIFF
--- a/roles/zuul/templates/etc/zuul/server-logging.conf
+++ b/roles/zuul/templates/etc/zuul/server-logging.conf
@@ -1,9 +1,9 @@
 # {{ ansible_managed }}
 [loggers]
-keys=root,zuul,gerrit,gear
+keys=root,zuul,gerrit,gear,requests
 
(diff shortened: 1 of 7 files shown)


roles/zuul/templates/etc/zuul/server-logging.conf
```diff
@@ -0,0 +1,63 @@
...
+args=('/var/log/zuul/server-debug.log',)
+
+[handler_normal]
+level=WARNING
+class=logging.handlers.WatchedFileHandler
+formatter=simple
+args=('/var/log/zuul/server.log',)
+
+[handler_requests]
+level=DEBUG
```
ConsoleCatzirl, line 48: Do we really want this at DEBUG level? This will be pretty noisy

omgjlk (approved):
Lets bring it in and than adjust verbosity as needed.

PR #229, opened 2017-03-14
Split nodepool role into v2 and v3

jamielennox:
In preperation for zuul v3 allow the playbook to deploy both.

### DIFF
--- a/roles/nodepool/tasks/nodepool_v3.yml
+++ b/roles/nodepool/tasks/nodepool_v3.yml
@@ -1,0 +1,36 @@
+---
+- name: Write logging config
+  template:
+    src: "etc/nodepool/logging.conf"
+    dest: "/etc/nodepool/{{ item }}_logging.conf"
+    owner: root
+    group: nodepool
+    mode: 0640
+  with_items:
+    - nodepool-builder
+    - nodepoold
+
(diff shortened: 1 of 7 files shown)


ConsoleCatzirl (commented):
:+1: Looks good to me, but I'll let others review as well

mlangbehn (commented):
Looks good to me.

roles/nodepool/tasks/nodepool_v3.yml
```diff
@@ -0,0 +1,36 @@
...
+- meta: flush_handlers
+
+- name: Enable and start systemd service
+  service:
+    enabled: yes
+    state: running
+    service: "{{ item }}"
+  with_items:
+    - nodepool-builder
+    - nodepoold
```
gandelman-a, line 36: It looks like the only thing that differs between nodepool_v2.yml  and nodepool_v3.yml is the services.  Could we have an internal '_nodepool_service_list' var that we use for with_items here, that is conditionally set to the correct services for the version of nodepol we're using? Then only have one nodepool.yml for both files

gandelman-a (approved):
This looks good. Gonna approve it in and we can start iterating on top of it in hoist proper, instead of our own forks.

PR #236, opened 2017-03-15
Remove shellcheck exclude from test-shellcheck.sh

mlangbehn:
Previously, we were excluding shellcheck rule 2046 in
tests/test-shellcheck.sh, rather than only excluding it for the line
that was causing the issue (tests/test-hoist-ansible.sh line 4). Now,
the exclude has been removed from test-shellcheck.sh.


ConsoleCatzirl (commented):
LGTM, but I'll let others review as well

omgjlk (approved):

PR #242, opened 2017-03-16
Add tools/vagrant-deploy.sh

mlangbehn:
Previously, vagrant deployments were failing because of an issue with
validate-ci.yml getting run before all of the nodes had been deployed.

To address this, we now have split the deploy steps into running the
ansible deployment, and then running the validation playbook. This can
be accomplished in one step by running the new tools/vagrant-deploy.sh
script.

Related-Issue: BonnyCI/projman#172

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -83,7 +83,6 @@ Vagrant.configure(2) do |config|
 
     zuul.trigger.after [:up, :provision] do
       run "vagrant ssh bastion -c 'sudo -i -u cideploy ssh -o StrictHostKeyChecking=no ubuntu@zuul.vagrant true'"
-      run "vagrant ssh bastion -c 'sudo -i -u cideploy /vagrant/tools/vagrant-run-ansible.sh --limit zuul'"
     end
 
     zuul.trigger.after :destroy do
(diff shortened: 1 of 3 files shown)


eggshell (commented):
lgtm, but others should review.

jasontclark (commented):
LGTM.

omgjlk (commented):
I have a reservation on this setup. From what I've gleaned on IRC, our vagrant setup is essentially starting a number of "machines" and each of those machines will have ansible ran on them, as a single host target, independent of any other host. This is a fundamental shift in deployment, and it means ordering of the playbooks cannot be relied upon. Our production runs execute the playbooks a single time with the full inventory at its disposal.  I'm wary of this difference sneaking up and biting us in unknown ways in the future, and making it difficult to validate our playbooks using the vagrant method.

If at all possible/efficient, I'd prefer if our vagrant made all the systems available, and a single playbook execution performed the configuration to make them a Bonny env, to more closely match our production deployment method.

I could be wrong, that this isn't possible, or this isn't a problem, as I'm a Vagrant novice. I won't leave a negative review here, others can offer their opinions and approve this anyway.

mlangbehn:
@j2sol: After our discussion on IRC, I agree with breaking the ansible run out to a second step. I will draft up my changes and update this pull request.

omgjlk (approved):
Looks good to me! I assume the docs will be updated too?

PR #245, opened 2017-03-16
Adding an account for jtclark

jasontclark:
Signed-off-by: Jason Clark <jason.clark@ibm.com>

### DIFF
--- a/create-cloud-resources.yml
+++ b/create-cloud-resources.yml
@@ -23,14 +23,14 @@
         - username: jkeating
           email: omgjlk@us.ibm.com
           isadmin: True
+        - username: jtclark
+          email: jtclark@linux.vnet.ibm.com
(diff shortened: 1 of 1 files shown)


create-cloud-resources.yml
```diff
@@ -29,6 +29,9 @@
         - username: rattboi
           email: bradon.kanyid@ibm.com
           isadmin: False
+        - username: jtclark
```
mlangbehn: We may want to move you up a bit so this list stays in alphabetical order?

mlangbehn (changes requested):
Looks almost great.

create-cloud-resources.yml
jasontclark: Sure thing

PR #247, opened 2017-03-17
Re-arranging the statement for order's sake.

jasontclark:


mlangbehn (commented):
Not blocking, because others may disagree, but I prefer pull requests only contain one commit (except where it only makes sense to add multiple commits).

Please:
- [ ] Squash commits

PR #251, opened 2017-03-17
Fix quartermaster service's working directory

rattboi:
Currently, the working directory for quartermaster is whatever
systemd says it is. It would be nice for plugins to be able to use
relative paths, so they don't need knowledge of where they are
installed. This changes the systemd service file to use the
quartermaster user's homedir as the working dir for the bot.


eggshell:
Awesome, thanks for addressing this! 👍

eggshell (commented):
Awesome, thanks for addressing this! 👍

mlangbehn (commented):
Looks good to me, but someone else should still review this.

SpamapS (approved):

PR #254, opened 2017-03-21
Open up nodepool default security group

gandelman-a:
The original spec only allowed SSh through to the nodepool slaves, which blocks
the telnet console ports.  This opens it up so at least bonnyci admins have
access.


omgjlk (approved):
LGTM. Is the security design documented anywhere outside of the spec?

SpamapS (approved):

PR #258, opened 2017-03-22
Reduce duplication, use two v2/v3 vars files instead.

jamielennox:


jamielennox:
missing signed-off etc, will let adam handle this i think

PR #259, opened 2017-03-22
Don't template when: conditions

jamielennox:
I'm never going to figure this quoting out. I thought you had to quote
in ansible v2, but it doesn't want to work that way.

### DIFF
--- a/roles/nodepool/tasks/main.yml
+++ b/roles/nodepool/tasks/main.yml
@@ -125,7 +125,7 @@
     - Restart nodepool
 
 - include: nodepool_v2.yml
-  when: "{{ not nodepool_zuul_v3 }}"
+  when: not (nodepool_zuul_v3 | bool)
 
 - include: nodepool_v3.yml
-  when: "{{ nodepool_zuul_v3 }}"
+  when: nodepool_zuul_v3 | bool


roles/nodepool/tasks/main.yml
```diff
@@ -125,7 +125,7 @@
     - Restart nodepool
 
 - include: nodepool_v2.yml
-  when: "{{ not nodepool_zuul_v3 }}"
+  when: not nodepool_zuul_v3
 
 - include: nodepool_v3.yml
-  when: "{{ nodepool_zuul_v3 }}"
+  when: nodepool_zuul_v3
```
omgjlk: I'd prefer these are run through the `bool` filter. Ansible/Jinja gets quirky with strings as "true" or "false".

omgjlk (dismissed):

roles/nodepool/tasks/main.yml
ConsoleCatzirl: I agree with @j2sol

mlangbehn: +1

anne-bonny:
Build failed.

- http://logs.bonnyci.org/check_github/BonnyCI/hoist/259/1490658167.6/bonnyci-run-check-multinode : FAILURE in 1m 12s

jamielennox:
recheck

omgjlk (approved):

PR #263, opened 2017-03-24
Bootstrap networks based on variables

jamielennox:
Remove the deploy specific variables from cloud-bootstrap defaults and
put them into the create-cloud-resources script.

### DIFF
--- a/create-cloud-resources.yml
+++ b/create-cloud-resources.yml
@@ -53,6 +53,26 @@
           interfaces:
             - control-plane-subnet
             - nodepool-subnet
+      security_groups:
+        - name: sg-control-plane
+          rules:
+            - remote_group: sg-control-plane
+              port_min: 1
+              port_max: 65535
+        - name: sg-zuul-merger
+          rules:
+            - remote_cidr: 10.0.0.0/8
+              port: 8858
+        - name: sg-ssh
+          rules:
+            - remote_cidr: 0.0.0.0/0
+              port: 22
+        - name: sg-http-https
+          rules:
+            - remote_cidr: 0.0.0.0/0
+              port: 80
+            - remote_cidr: 0.0.0.0/0
+              port: 443
       servers:
         - name: bastion
           floating_ip_address: 169.44.171.83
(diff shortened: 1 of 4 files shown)


roles/cloud-bootstrap/tasks/network.yml
```diff
@@ -1,56 +1,34 @@
...
-        - "{{ control_plane_subnet.subnet.id }}"
-        - "{{ nodepool_subnet.subnet.id }}"
-    project: "{{ prod_project }}"
+    name: "{{ item.name }}"
+    network: "{{ item.network | default(omit) }}"
+    interfaces: "{{ item.interfaces | default(omit) }}"
+    project: "{{ item.project }}"
     auth:
-      project_name: "{{ prod_project }}"
+      project_name: "{{ item.project }}"
```
mlangbehn: Did you mean to have this here:

```yaml
      project_name: "{{ item.project }}"
  with_items: "{{ routers }}"
```

mlangbehn (commented):
Quick inline question.

jamielennox:
There's a lot more to be done to this role - but this appears to work so long as you specify project: to the network modules.

create-cloud-resources.yml
```diff
@@ -38,6 +38,21 @@
...
+          shared: True
+          project: bonnyci
+          cidr: 10.0.0.0/8
+      routers:
+        - name: bonnyci-router
+          network: external
+          project: bonnyci
+          interfaces:
+            - control-plane-subnet
+            - nodepool-subnet
```
gandelman-a, line 55: so if we stop using the project name variables and specify here, we'd need to do that for everything specified in this file. seems like it'd be easier to set as a var, have it used in the roles' tasks so it can DTRT?

gandelman-a:
I think this can be closed now?

mlangbehn:
@jamielennox, do we still need this change? If we do, please resolve the merge conflicts. Maybe squash your commits, too?

jamielennox:
Abandoning. We went the simpler way for now and just encoded everything in the role.

PR #264, opened 2017-03-24
Ssh rekey

ConsoleCatzirl:


roles/bastion/defaults/main.yml
```diff
@@ -12,11 +15,8 @@ bastion_users:
   - rmoe
   - bradonk
   - doc
-  - cideploy
   - jkeating
   - jesusaur
```
rattboi: You're down here in remove_users. Are we doing a follow-up commit to add back those users who need it to bastion_users?

rattboi (changes requested):

rattboi (approved):

gandelman-a (approved):

PR #270, opened 2017-03-27
Adding a new bastion key for the jason_clark account.

jasontclark:
Signed-off-by: Jason Clark <jason.clark@ibm.com>


roles/bastion/defaults/main.yml
```diff
@@ -6,13 +6,13 @@ bastion_users:
   - cideploy
   - jesusaur
   - olaph
+  - jason_clark
```
loafyloaf: I believe the idea is to keep this list alphabetized

PR #275, opened 2017-03-28
 Slice up provisioning to support multiple zuul environments per …

gandelman-a:
…bastion

This creates a common network and puts services that are shared among
environments on it, including the bastion.

It then moves per-environment hosts in to an 'environments' list that
will allow easily expanding our infra to include many zuul environments.

A reasonable set of defaults are used so that individual users can spin up
similar environments in their tenant space.  Those may be overriden to supply
site-specific values for production.

### DIFF
--- a/create-cloud-resources.yml
+++ b/create-cloud-resources.yml
@@ -38,45 +38,77 @@
         - username: rattboi
           email: bradon.kanyid@ibm.com
           isadmin: False
-      servers:
+      common_servers:
         - name: bastion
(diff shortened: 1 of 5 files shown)


jamielennox (approved):

anne-bonny:
Merge Failed.

This change or one of its cross-repo dependencies was unable to be automatically merged with the current state of its repository. Please rebase the change and upload a new patchset.

PR #276, opened 2017-03-29
separate datadog-builder from bastion role

loafyloaf:
Signed-off-by: matthew wagoner <zxkuqyb@gmail.com>

### DIFF
--- a/dd-builder.yml
+++ b/dd-builder.yml
@@ -1,0 +1,19 @@
+- name: Configure datadog-builder
+  hosts: bastion
+  become: true
+
+  roles:
+    - role: datadog-builder
+      datadog_builder_secrets:
+        api_key: "{{ secrets.datadog.api_key }}"
+        app_key: "{{ secrets.datadog.ansible_app_key }}"
+    - role: logrotate
+      logrotate_configs:
+        - name: datadog-builder
+          path: /var/log/datadog-builder/*log
+          options:
+            - compress
+            - missingok
+            - rotate 30
+            - daily
+            - notifempty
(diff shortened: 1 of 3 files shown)


dd-builder.yml
```diff
@@ -0,0 +1,19 @@
...
+    - role: logrotate
+      logrotate_configs:
+        - name: datadog-builder
+          path: /var/log/datadog-builder/*log
+          options:
+            - compress
+            - missingok
+            - rotate 30
+            - daily
+            - notifempty
```
gandelman-a, line 19: So as-is, nothing is setup to actually run this playbook in the same way we  do bastion.yml or install-ci.yml.  Was this intentional?

gandelman-a:
Without a commit msg its kinda hard to determine the intent of this?  Its just moving the datadog-builder role off of the bastion into its own group, but results in that stuff not running anymore.

gandelman-a (changes requested):

loafyloaf:
so, at the time, datadog-builder was running on two bastions - contrasjc and opentechsjc.  While this shouldn't really be a problem, the monitors.yml file was not getting updated on contrasjc, so when 'new' bastion created a monitor, 'old' bastion would delete it.  This is a bit moot now, seeing as how contrasjc is gone.  But perhaps we may get a second cloud in the future...

(aplogies for the lacking commit msg)

dd-builder.yml
ConsoleCatzirl: Is the idea to create a new ansible-runner cron for this playbook?

mlangbehn:
Per discussion with @loafyloaf in IRC, this pull request is no longer needed. Closing accordingly.

PR #277, opened 2017-03-29
Manage multiple NP projects, manage production quotas

gandelman-a:
NOTE this is based ontop of changes in another PR: https://github.com/BonnyCI/hoist/pull/275

This updates account management to allow creation of multiple nodepool
projects to be used across multiple zuul/nodepool environments.

It also adds the ability to set quotas for production and nodepool
accounts, defaulting to opening them up to unlimited usage.  It
vendors a new, unreleased os_quota.py module in library/ which includes
some fixes to a bug we hit when trying to use it on our cloud.

This also fixes a typo in a defaults file.

Related-Issue: BonnyCI/projman#139
Closes-Issue: BonnyCI/projman#194

Signed-off-by: Adam Gandelman <adamg@ubuntu.com

### DIFF
--- a/library/os_quota.py
+++ b/library/os_quota.py
@@ -1,0 +1,506 @@
+#!/usr/bin/python
+# Copyright (c) 2016 Pason System Corporation
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+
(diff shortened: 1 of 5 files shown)


roles/cloud-bootstrap/defaults/main.yml
```diff
@@ -67,7 +67,7 @@ environment_security_group_rules:
       - remote_cidr: "{{ nodepool_network_cidr }}"
         port: 8858
 
-ommon_servers:
+common_servers:
```
SpamapS, line 103: The commit message on this doesn't quite match the content. :-P

SpamapS (dismissed):
LGTM but needs to be squashed maybe?

gandelman-a:
Whoops. Removed the stale commit and updated commit msg about the typo fix.

library/os_quota.py
```diff
@@ -0,0 +1,506 @@
+#!/usr/bin/python
+# Copyright (c) 2016 Pason System Corporation
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
```
omgjlk, line 5: So we don't account for files we've vendored in that are in a license other than Apache. Since we have a few files like this, we need to change the LICENSE or README.rst files to point out the files that are not our normal APACHE license. ( should not block this PR )

omgjlk (approved):
Looks good to me. Needs a follow up to address licensing issues.

PR #278, opened 2017-03-30
Begin expanding inventory for new V3 environment

gandelman-a:
This checks in some new inventory files for the V3 env.

It also hacks up the ansible-runner role to allow us to manage
any number of runner tasks for differente environments, instead
of relying on a global variable.


anne-bonny:
Merge Failed.

This change or one of its cross-repo dependencies was unable to be automatically merged with the current state of its repository. Please rebase the change and upload a new patchset.

PR #279, opened 2017-03-30
Add clint's SSH keys to bastion

SpamapS:
Signed-off-by: Clint Byrum <clint@fewbar.com>


gandelman-a (changes requested):
The SSH key looks like the same one that was in the repo before we did the rekey'ing. Can you generate a new pair instead?

SpamapS:
So this is really confusing. I have one SSH key on my work machine. I did go ahead and rotate it, since the implication is that perhaps someone compromised it and used that to attack our servers.

However, I'm just going to keep from having SSH, so closing this w/o updating it for now.

PR #281, opened 2017-03-30
Update inventory

gandelman-a:


inventory/group_vars/vagrant
```diff
@@ -13,8 +13,6 @@ elasticsearch_jvm_heap_size: 250m
 filebeat_output_logstash_hosts:
   - elk.vagrant:5044
 
-hoist_repo: /vagrant/.git
```
mlangbehn, line 16: This line also appears in `inventroy/group_vars/all` line 8:

```yaml
hoist_repo: https://github.com/BonnyCI/hoist.git
```

With `hoist_repo` removed from `bastion.yml`, I believe we no longer need that reference either?

roles/ansible-runner/tasks/runner.yml
```diff
@@ -0,0 +1,49 @@
...
+    repo: "{{ item.repo }}"
+    dest: "/opt/source/{{ item.name }}"
+    update: true
+
+- name: Install config file
+  template:
+    src: ansible-runner-config
+    dest: /etc/default/{{ item.name }}
+    mode: 0600
+    owner: "{{ item.user }}"
```
mlangbehn, line 12: Is there a reason we are starting to work in `/opt/source/`, and then moving to `/etc/default/`, and then back to `/opt/source/`? I would suggest either moving this block up to the beginning, or after `Install playbook deps`.

roles/ansible-runner/templates/ansible-runner-config
```diff
@@ -1,3 +1,3 @@
-ANSIBLE_PLAYBOOK={{ ansible_playbook }}
-ANSIBLE_INVENTORY={{ ansible_inventory }}
-ANSIBLE_SSH_USER={{ ansible_remote_user | default('') }}
+ANSIBLE_PLAYBOOK={{ item.playbook }}
+ANSIBLE_INVENTORY=/opt/source/{{ item.name }}/inventory/{{ item.inventory }}
+ANSIBLE_SSH_USER={{ item.user | default('')}}
```
mlangbehn, line 3: Shouldn't this be:

```jinja2
ANSIBLE_SSH_USER={{ item.user | default('') }}
```

roles/ansible-runner/tasks/runner.yml
```diff
@@ -0,0 +1,49 @@
...
+    requirements: "/opt/source/{{ item.name }}/requirements.txt"
+  when: requirements_txt.stat.exists
+
+- name: Ensure cron log path
+  file:
+    state: directory
+    path: "/var/www/html/cron-logs/{{ item.name }}"
+    mode: 0755
+    owner: "{{ item.user }}"
+    group: www-data
```
mlangbehn, line 37: Should this go after the `include: secret-file.yml`? Just to help keep the tasks organized.

inventory/host_vars/bastion.vagrant
```diff
@@ -1,4 +1,15 @@
...
 
-ci_inventory: vagrant
+ansible_runner_tasks:
+    - name: system-ansible
+      playbook: bastion.yml
+      inventory: vagrant
+      repo: /vagrant/.git
+      user: root
+
+    - name: cideploy-v2
```
mlangbehn, line 11: Should this be?:

```yaml
    - name: cideploy
```

mlangbehn (commented):
Please see inline comments.

roles/ansible-runner/tasks/runner.yml
gandelman-a: This was copied/pasetd from main.yml, i suppose the ordering just happened?  Will update here, tho

roles/ansible-runner/tasks/runner.yml
gandelman-a: ditto ^

PR #282, opened 2017-03-30
Fix zuulv3 logging configs, maintain 2 zuul.conf templates

gandelman-a:
This fixes some issues with zuul v3 logging and condenses all logging
config templates into one.

It also splits the zuul.conf template into two version specific templates.
There's already enough in-line flow control, adding additional v2/v3
conditionals makes it difficult to manage.


mlangbehn (commented):
Looks good to me, but someone else should review.

ConsoleCatzirl (approved):

PR #283, opened 2017-03-31
Begin expanding inventory for new V3 environment

gandelman-a:
This checks in some new inventory files for the V3 env.

Since we can no longer rely on the group membership to determine which zuul_components to run, this setting is moved to host_vars for the merger, and the opentech-* group_vars for the others.


inventory/group_vars/opentech-sjc
```diff
@@ -41,4 +41,8 @@ nodepool_zmq_publishers:
 zuul_gearman_server: zuul.internal.opentechsjc.bonnyci.org
 zuul_logs_url: http://logs.bonnyci.org
 
+zuul_components:
```
ConsoleCatzirl: By putting this here, any opentech-sjc host that runs the zuul role will get zuul launchers and servers, but we only want one server per deployment.

inventory/group_vars/mergers
```diff
@@ -18,6 +18,3 @@ bonnyci_zuul_merger_apache_vhosts:
       ScriptAlias /p/ /usr/lib/git-core/git-http-backend/
 
 bonnyci_zuul_merger_git_dir: /var/lib/zuul/git
-
-zuul_components:
```
ConsoleCatzirl, line 22: Why move this from group_vars to host_vars?

inventory/host_vars/merger01.internal.opentechsjc.bonnyci.org
```diff
@@ -0,0 +1,3 @@
+---
+zuul_components:
```
ConsoleCatzirl, line 2: Moving this from group_vars to host_vars makes it easy to accidentally deploy a new zuul server instead of a new zuul merger when trying to add additional mergers.

ConsoleCatzirl (dismissed):
Most of this change looks good, but I worry about moving the zuul_components around. When it comes time to add another zuul merger, this makes it easy for us to accidentally deploy a new server and launcher instead of a new merger if we forget to add a new host_vars file.

inventory/host_vars/merger01.internal.opentechsjc.bonnyci.org
gandelman-a: So we discussed on IRC and decided to manage everything in host_vars, rather than a potentially confusing combo of host_vars and group_vars.

ConsoleCatzirl (commented):
Looks good to me, but will others review as well

inventory/group_vars/mergers
jamielennox: I don't think it's a problem here - but conceptually lists don't merge when they are overridden. If you were to put a merger on the same host as something else the zuul_components var would be overridden.

inventory/group_vars/opentech-sjc-v3
```diff
@@ -0,0 +1,66 @@
...
+
+zuul_gearman_server: zuul.internal.v3.opentechsjc.bonnyci.org
+zuul_logs_url: http://logs.bonnyci.org
+
+zuul_connections:
+  gerrithub:
+    driver: gerrit
+    user: bonnyci
+    server: review.gerrithub.io
+    sshkey: /var/lib/zuul/.ssh/id_rsa
```
jamielennox, line 55: does this user exist on gerrithub? we need to upload this key.

jamielennox (approved):
Let's try it and see.

PR #292, opened 2017-04-04
Drop Zuul V3 multinode label, decrease min-ready, update provider

gandelman-a:
Our cloud is running out of resources. We don't initially need any resources
for our V3 testing.  This drops the min-ready to 1, drops the multinode
label entirely and uses a smaller instance size for the slaves.

It also updates the node provider name for the V3 environment, which gets set
in slave instances' metadata and is used by nodepool to filter instances.  This
should be unique.


eggshell (commented):
looks good to me.

omgjlk (approved):

PR #297, opened 2017-04-05
Nodepool-v3 network should create nodes with open access

jamielennox:
Nodepool will try to ssh to the newly created nodes to verify that they
are up. To do this they need to be in a security group that allows it.
The way this is handled in v2 is that the default sg just allows access
to all ports from the control plane network. This is probably
unnecessary and should be just SSH, but replicate it for now.

This has not actually been tested because I don't really have the
ability, but it reflects a change I had to make on the cloud.

### DIFF
--- a/roles/cloud-bootstrap/defaults/main.yml
+++ b/roles/cloud-bootstrap/defaults/main.yml
@@ -46,6 +46,7 @@ prod_quota:
 prod_nodepool_projects:
     - name: nodepool
       user: nodepool
+      controlplane_network_cidr: 192.168.10.0/24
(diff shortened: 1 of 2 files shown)


anne-bonny:
Build failed.

- http://logs.bonnyci.org/check_github/BonnyCI/hoist/297/1491356892.45/bonnyci-run-check-multinode : POST_FAILURE in 15s

omgjlk:
recheck

omgjlk (approved):

gandelman-a:
recheck

PR #304, opened 2017-04-05
Drop elk and logs hosts from opentech-sjc

gandelman-a:
These hosts now live in the opentech-sjc-common inventory.  They
can be removed from opentech-sjc to avoid deploying to them twice.


loafyloaf (approved):
makes sense to me

PR #307, opened 2017-04-05
Add a default deny apache vhost to elk

gandelman-a:
A previous commit that was meant to setup kibana reverse proxying
mistakenly opened the host up as a public forward proxy, and was
picked up as a backend for a public proxying service. Forward
proxying was disabled, but we're still getting tons of incoming
proxy requests.  This sets up a default vhost to deny these and
only accept requests explicitly bound for configured vhosts.  This
will hopefully result in the address being pulled from the proxy
service's database.

### DIFF
--- a/inventory/group_vars/monitoring
+++ b/inventory/group_vars/monitoring
@@ -2,7 +2,16 @@ bonnyci_kibana_apache_mods_enabled:
   - proxy.load
   - proxy_http.load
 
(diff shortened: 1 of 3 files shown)


loafyloaf (commented):
oh, nice.  very yes

omgjlk (approved):

PR #313, opened 2017-04-06
Add ssh-host-keys role

jamielennox:
We need to have the SSH host keys for logs on the zuul servers so that
they can upload logs to them. The most generic way i can think of is to
make a role that fetches and saves the host keys for all the specified
groups.

### DIFF
--- a/install-ci.yml
+++ b/install-ci.yml
@@ -85,6 +85,9 @@
       tags:
         - monitoring
 
+    - role: ssh-host-keys
+      ssh_host_keys_hostnames: log
+
     - role: apache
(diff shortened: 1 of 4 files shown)


gandelman-a (changes requested):
This looks cool but I'm not sure it will work.  The logs host is part of the opentech-sjc-common inventory and its hostvars wont be available to the zuul play, which is run with either opentech-sjc or opentech-sjc-v3 inventories.  We could potentially add the logs server back to the non-common inventories so the hostvars are available, but skip running them during the non-common playbooks with a '--skip-tags logs' ?

jamielennox:
Yea, i think you're right - it only works in my env based on the way it's set up.

PR #318, opened 2017-04-07
Add force option to ansible-runner

mlangbehn:
Previously, if ansible-runner needed to be run by hand but was disabled
to stop the cronjob from running every 15 minutes, one had to reenable
ansible-runner, thus permitting the cronjob to run at it's next
interval. This opened unwanted possibilities like the cronjob running
in the background when a manual run was wanted, or forgetting to disable
the job again.

Now there is a force option (`--force`) for ansible-runner.
Unfortunately, the standard short form (`-f`) will not be available to
us, because we need to be able to pass additional options to ansible,
which uses `-f` to specify a number of forks.

Related-Issue: BonnyCI/projman#195


eggshell (commented):
looks good to me. will leave open for others to review.

omgjlk (approved):

PR #320, opened 2017-04-07
Fix handling of force option in ansible-runner

mlangbehn:
Previously, the --force option was being passed to ansible after being
used to force the ansible-runner script because ansible uses the
deprecated optparse, which allows for abbreviated options. As a result
`--force` was being interpreted by ansible-playbook as
`--force-handlers`.

Now, ansible-runner uses the more obvious `--force-runner` (per
@gandelman-a's request), which switches on `force_runner`. All other
arguments are moved to the `ansible_options` string, which is in turn
passed to `ansible-playbook`.

With these changes made, run-one breaks the script. However, our
deployments are currently unlikely to be impacted by us not using
run-one. As such, it has been removed.

Related-Issue: BonnyCI/projman#211

### DIFF
--- a/roles/ansible-runner/files/usr/local/bin/ansible-runner
+++ b/roles/ansible-runner/files/usr/local/bin/ansible-runner
@@ -2,7 +2,7 @@
 # Runs system wide ansible configured in /etc
 
 env=$1
-
+shift
 if [ -f /etc/default/"$env" ]; then
   . /etc/default/"$env"
 fi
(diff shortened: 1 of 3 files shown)


roles/ansible-runner/files/usr/local/bin/ansible-runner
```diff
@@ -11,29 +11,33 @@ envs_root=${ANSIBLE_RUNNER_ENV_ROOT:-/opt/source}
...
 
-if [[ "$*" == *"--force"* && "$*" != *"--force-"* ]]; then
-    echo "--force detected, ignoring /etc/disable-ansible-runner-$env..."
+for argument in "$@"; do
+    if [[ "$argument" == "--force-runner" ]]; then
+        force_runner=true
+    else
+        ansible_options="$ansible_options $argument"
+    fi
+    shift
```
gandelman-a: This 'shift' is not needed since we're not actually handling positional arguments /w the numerical variables at this point ($1, $2, etc)

gandelman-a (changes requested):

roles/ansible-runner/files/usr/local/bin/ansible-runner
mlangbehn: True, that is left over from testing.

ConsoleCatzirl (approved):

gandelman-a (approved):

omgjlk:
gate_recheck

omgjlk:
gate-recheck

PR #330, opened 2017-04-11
Merge ARA and tailon vhosts

jamielennox:
Having two vhosts on port 80 results in them getting merged in some
unusual ways. It for example prevents the .log file being opened in a
browser - but not it being available.

Given they are defined right next to each other just merge them into one
vhost.


mlangbehn (commented):
Looks good to me.

eggshell (approved):

PR #331, opened 2017-04-11
Add bonnyci.org link to merge-failure-message

eggshell:
We should direct users to a place with some documentation on where to
get help on resolving merge failures.

Implements: BonnyCI/projman#199
Depends-On: BonnyCI/bonnyci.org#57

### DIFF
--- a/roles/zuul/templates/etc/zuul/config/layout.yaml
+++ b/roles/zuul/templates/etc/zuul/config/layout.yaml
@@ -7,6 +7,7 @@ pipelines:
   - name: check_github
     manager: IndependentPipelineManager
     source: github
+    merge-failure-message: "{{ zuul_merge_failure_message }}"
     trigger:
       github:
         - event:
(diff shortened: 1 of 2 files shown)


roles/zuul/templates/etc/zuul/config/layout.yaml
```diff
@@ -33,6 +33,7 @@ pipelines:
     manager: DependentPipelineManager
     precedence: high
     source: github
+    merge-failure-message: "{{ zuul_merge_failure_message }}"
```
ConsoleCatzirl, line 37: Let's also set this in the check_github pipeline

ConsoleCatzirl (changes requested):
:+1: Looks good, but let's set it in both pipelines.

roles/zuul/templates/etc/zuul/config/layout.yaml
eggshell: @jesusaurus done.

ConsoleCatzirl (approved):

PR #339, opened 2017-04-13
Reduce min-ready to 1 for both labels

ConsoleCatzirl:
Since we have a small cloud quota, it's easy for the ubuntu-hoist label
to take up all of our resources since it uses multiple nodes, leaving us
with no ubuntu-xenial nodes ready. Drop both min-ready counts to 1 to make
it more likely that we will ready nodes of both labels.

This means we may spend more time waiting for nodes to build.

### DIFF
--- a/inventory/group_vars/opentech-sjc
+++ b/inventory/group_vars/opentech-sjc
@@ -35,12 +35,12 @@ nodepool_providers:
 nodepool_labels:
   - name: ubuntu-xenial
     image: ubuntu-xenial
-    min-ready: 3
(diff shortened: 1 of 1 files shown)


anne-bonny:
Build failed.

- http://logs.bonnyci.org/check_github/BonnyCI/hoist/339/1492107324.4/bonnyci-run-check-multinode : FAILURE in 4m 26s

gandelman-a (changes requested):
I think this can be closed. We went the other route of actually doubling our max-servers instead.

PR #341, opened 2017-04-14
Add script to perform dib image validation

eggshell:
We should check for some necessities during the image build process and
fail if any of them are not found or non-functional. This will add a
script that runs late in the build process and checks:

  * python 2.7.x installation
  * virtualenv installation
  * zuul-cloner check

As it stands right now, the checks are fairly rudimentary and just
ensure that the commands run without error.

Implements: BonnyCI/projman#215

### DIFF
--- a/roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
+++ b/roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+function reporter () {
+  message="$1"
+  shift
+  echo
+  echo "$message"
+  for (( i=0; i<${#message}; i++ )); do
+    echo -n '-'
+  done
+  echo
+}
+
+function build_failure () {
+  reporter "Error encountered! Exiting build."
+  reporter "$1" >&2
+  exit 1
+}
+
+function check_python () {
+  reporter "Validating python installation."
+  py_ver=$(python --version 2>&1)
+  if [[ ! "$py_ver" == *"Python 2.7"* && ! "$py_ver" == *"Python 3"* ]]; then
+    build_failure "Python not found!"
+  fi
+}
+
+function check_virtualenv () {
+  reporter "Validating virtualenv installation and use."
+  if [[ ! $(which virtualenv) ]]; then
+    build_failure "virtualenv not found!"
+  fi
+
+  if [[ ! $(mkdir /tmp/test-venv && virtualenv /tmp/test-venv) ]]; then
+    build failure "virtualenv not functional!"
+  else
+    rm -rf /tmp/test-venv
+  fi
+}
+
+function check_zuul_cloner () {
+  reporter "Validating zuul-cloner installation."
+  if [[ ! $(zuul-cloner -h) ]]; then
+    build_failure "zuul-cloner not found or not functional!"
+  fi
+}
+
+function main () {
+  check_python
+  check_virtualenv
+  check_zuul_cloner
+}
+
+main "$@"


roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
```diff
@@ -0,0 +1,50 @@
...
+  reporter "Validating python installation."
+  python_version=$(python --version)
+  if [[ ! "$python_version" == *"Python 2.7" ]]; then
+    build_failure "Python does not exist or is not version 2.7.x"
+  fi
+}
+
+function check_virtualenv () {
+  reporter "Validating virtualenv installation."
+  which virtualenv
```
gandelman-a: maybe we could create a virtualenv in /tmp just to confirm the executable is doing the right thing?

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
```diff
@@ -0,0 +1,50 @@
...
+  reporter "${message}" >&2
+  exit 1
+}
+
+function check_python () {
+  reporter "Validating python installation."
+  python_version=$(python --version)
+  if [[ ! "$python_version" == *"Python 2.7" ]]; then
+    build_failure "Python does not exist or is not version 2.7.x"
+  fi
```
gandelman-a: not sure we want to fail here if we have a non-py27 version.  we may at some point want to provide an image with py3 as the default? im not sure..

gandelman-a (dismissed):
Looks good tho I'm not sure how / what we want to validate about python. We may want to offer a py3-based image, or build from newer distro versions that ship py3 as the default.   Do we even want to validate the python version, or just assume a functional python if the zuul-cloner test passes?

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
eggshell: @gandelman-a good idea, went ahead and added a check for that. thanks!

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
eggshell: @gandelman-a I wasn't sure how this would play out in the future, so just went with what we need now. I can easily change it to a "do we have any python?" check.

omgjlk (approved):

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+function reporter () {
+  message="$1"
+  shift
```
mlangbehn: Do we need to set message and shift? Why not just replace this bit with this?

```shell
    echo -e "\n$1\n"
```

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
```diff
@@ -0,0 +1,57 @@
...
+  echo
+  echo "${message}"
+  for (( i=0; i<${message}; i++ )); do
+    echo -n '-'
+  done
```
mlangbehn: What are we trying to do with this for loop? Since `$message` is not an integer, I would expect this to fail.

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
```diff
@@ -0,0 +1,57 @@
...
+  python_version=$(python --version)
+  if [[ ! "$python_version" == *"Python 2.7" ]]; then
+    build_failure "Python does not exist or is not version 2.7.x"
+  fi
+}
+
+function check_virtualenv () {
+  reporter "Validating virtualenv installation and use."
+  which virtualenv
+  if [[ "$?" -ne 0 ]]; then
```
mlangbehn: This can be simplified to:

```shell
  if [[ ! $(which virtualenv) ]]; then
```

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
```diff
@@ -0,0 +1,57 @@
...
+    build_failure "virtualenv not found"
+  fi
+
+  mkdir /tmp/test-venv && virtualenv /tmp/test-venv
+  if [[ "$?" -ne 0 ]]; then
```
mlangbehn: This can be simplified to:

```shell
  if [[ ! $(mkdir /tmp/test-venv && virtualenv /tmp/test-venv) ]]; then
```

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
```diff
@@ -0,0 +1,57 @@
...
+    build failure "virtualenv not functional"
+  else
+    rm -rf /tmp/test-venv
+  fi
+}
+
+function check_zuul_cloner () {
+  reporter "Validating zuul-cloner installation."
+  zuul-cloner -h
+  if [[ "$?" -ne 0 ]]; then
```
mlangbehn: This can be simplified to:

```shell
  if [[ ! $(zuul-cloner -h) ]]; then
```

mlangbehn (commented):
Please review comments in file.

mlangbehn:
Note, if you run `shellcheck` against this, it will fail right now. The reason it is not being run in CI is that the file name does not include the `.sh` extension.

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
```diff
@@ -0,0 +1,57 @@
...
+  echo "${message}"
+  for (( i=0; i<${message}; i++ )); do
+    echo -n '-'
+  done
+  echo
+}
+
+function build_failure () {
+  reporter "Error encountered! Exiting build."
+  reporter "${message}" >&2
```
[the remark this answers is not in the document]
eggshell: good catch, should be "$1"

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
eggshell: fixed.

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
eggshell: fixed.

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
eggshell: fixed.

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
eggshell: @mlangbehn This is a function I've used in bash scripts for years. It interprets the length of the message and prints the correct number of dashes. So if your message was "Installing python", it would look like this:

```
Installing python
-----------------
```

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
```diff
@@ -0,0 +1,53 @@
...
+
+function build_failure () {
+  reporter "Error encountered! Exiting build."
+  reporter "$1" >&2
+  exit 1
+}
+
+function check_python () {
+  reporter "Validating python installation."
+  if [[ ! $(python --version) == *"Python 2.7"* ]]; then
```
mlangbehn: Fun fact! The output of `python --version` is actually sent to stderr (I'm not sure why).

This can be seen by running the following:

```shell
$ python --version | cut -d ' ' -f2
Python 2.7.13
$ python --version 2>&1 | cut -d ' ' -f2
2.7.13
```

As such, we actually want this line to be:

```shell
  if [[ ! "$(python --version 2>&1)" == *"Python 2.7"* ]]; then
```

**NOTE:** without this change, the script will always fail on `check_python`.

eggshell: @mlangbehn yep, fixed it.

roles/nodepool/files/etc/nodepool/elements/nodepool/cleanup.d/90-validate-dib-image
eggshell: @gandelman-a so I did some ruminating on this and modified the conditional to include a check for 3.x wildcards.

mlangbehn (approved):
Looks great! Thanks, @eggshell!

eggshell:
gate_recheck

mlangbehn:
gate_recheck

mlangbehn:
recheck

PR #346, opened 2017-04-17
Fix formatting issue with new quartermaster config

rattboi:


omgjlk (approved):
ALL THE NICE YAML!

PR #349, opened 2017-04-17
Replace anne-bonny user with bonnyci[bot] in layout

jamielennox:
When using the integration our CI user is going to be bonnyci[bot]
instead of anne-bonny. With the dependant transition patch we are faking
the anne-bonny user into providing bonnyci[bot] reviews, however this
means we need to have our layout reflect the new user.

Depends-On: BonnyCI/zuul#39


omgjlk (dismissed):
I like this, but leaving a -1 until we restart and move some accounts over.

PR #351, opened 2017-04-18
Add a way to disable the ansible-runners from ansible

jamielennox:
Add a disable flag that disables the ansible-runner from running on
cron. This is useful for testing where I don't want all the runners to
actually start.

I haven't added the reverse that removes the flag from git as I feel you
probably want to control this manually so the automation doesn't remove
your manual hold.

### DIFF
--- a/roles/ansible-runner/tasks/runner.yml
+++ b/roles/ansible-runner/tasks/runner.yml
@@ -33,6 +33,12 @@
   secret_file_path: "{{ ansible_runner_secrets_path }}"
   secret_file_user: "{{ item.user }}"
 
+- name: Disable cron job
+  copy:
+    content: ""
+    dest: "/etc/disable-ansible-runner-{{ item.name }}"
+  when: item.disabled | default(ansible_runner_disabled)
+
 - name: Ensure cron log path
   file:
     state: directory
--- a/roles/ansible-runner/defaults/main.yml
+++ b/roles/ansible-runner/defaults/main.yml
@@ -1,6 +1,7 @@
 ansible_runner_user: root
 ansible_runner_secrets_path: /etc/secrets.yml
 ansible_runner_virtualenv: /opt/ansible
+ansible_runner_disabled: False
 datadog_callback_url: https://raw.githubusercontent.com/DataDog/ansible-datadog-callback/master/datadog_callback.py
 
 # Example:


jamielennox:
recheck

roles/ansible-runner/tasks/runner.yml
```diff
@@ -33,6 +33,12 @@
   secret_file_path: "{{ ansible_runner_secrets_path }}"
   secret_file_user: "{{ item.user }}"
 
+- name: Disable cron job
+  copy:
```
ConsoleCatzirl, line 37: (nit) I would prefer to use the file module with `state: touch` instead of the copy module.

(-1) Let's also remove the file when the runner is not disabled

ConsoleCatzirl (dismissed):

roles/ansible-runner/tasks/runner.yml
jamielennox: state: touch will set a "changed" on every run, i though this way would be more idempotent.

The problem with removing the file when the runner is disabled is that if you log in to the box and touch the file to manually stop running the ci then on the next run the ci will remove it.

jamielennox: The way i wanted to use this was to do -e ansible_runner_disabled=True from the CLI to turn everything off in development.

ConsoleCatzirl: Oh, I didn't realize that state: touch would always report changed.

ConsoleCatzirl: If we are going to support disabling specific runners with a change hoist to set item.disabled: True, then I think reverting that commit should automatically re-enable the runner (as long as the disabled runner isn't the bastion playbook, since that would never run to re-enable the runner).

And if we remove the disable file here when it's not set to disabled, your use case of setting -e ansible_runner_disabled=True from the CLI still works since that would disable the bastion playbook runner, and you would need to manually remove that disable file to re-enable the runners.

jamielennox:
recheck

jamielennox:
recheck

jamielennox:
recheck

jamielennox:
recheck

jamielennox:
recheck

jamielennox:
recheck

jamielennox:
recheck

jamielennox:
recheck

omgjlk (approved):
I think this is fine as it is.

rattboi:
recheck

ConsoleCatzirl:
gate-recheck

PR #352, opened 2017-04-18
Split ansible-runner outputs into bastion host

jamielennox:
The ansible-runner role should just be setting up how to configure the
ansible-runner. Setting up the apache, ara, uwsgi and other pieces
should be bastion specific.

Split these dependencies away from the ansible-runner.


jamielennox:
recheck

rattboi (approved):
LGTM :+1:

PR #353, opened 2017-04-18
We need to run nodepool as the nodepool user

ConsoleCatzirl:
We can't access the clouds.yaml secrets unless we run the command as
the nodepool user.


jamielennox (changes requested):
Sorry, should have realized before - we don't want to build the image from dib, we just need to delete some and let them recreate.

PR #354, opened 2017-04-18
Add dib image build test to hoist

eggshell:
Currently, we have no tests for dib image builds using the elements
defined in hoist. This results in untested image builds ending up
in prod. This will add a job for that.

Implements: BonnyCI/projman#219

### DIFF
--- a/roles/zuul/files/jobs/dib.yaml
+++ b/roles/zuul/files/jobs/dib.yaml
@@ -1,0 +1,26 @@
+- job:
+    name: 'dib-build'
+    node: 'ubuntu-xenial'
+    builders:
+      shell: |
+        virtualenv $HOME/dib-build
+        source $HOME/dib-build/bin/activate
+        pip install diskimage-builder
+
+        sudo apt-get install -y debootstrap qemu-utils curl \
+                                uuid-runtime parted kpartx
+
+        sudo git clone https://github.com/BonnyCI/hoist.git /opt/hoist
+        export ELEMENTS_PATH=/opt/hoist/roles/nodepool/files/etc/nodepool/elements/nodepool
+        # NOTE: be sure the list of nodepool elements here stays consistent
+        # with the elements specified in the nodepool role.
+        sudo disk-image-create -x -t qcow2 --checksum --no-tmpfs            \
+                          --qemu-img-options 'compat=0.10'                  \
+                          -o /opt/nodepool/images/ubuntu-xenial             \
+                          ubuntu-minimal vm simple-init growroot            \
+                          openssh-server devuser haveged pip-and-virtualenv \
+                          nodepool bonnyci-nodepool
+
+    publishers:
+      - console-log
+      - bonnyci-logs
--- a/roles/zuul/templates/etc/zuul/config/layout.yaml
+++ b/roles/zuul/templates/etc/zuul/config/layout.yaml
@@ -69,6 +69,8 @@ jobs:
   # every job invokes the set_node_options file in bonnyci_functions.py
   - name: ^.*$
     parameter-function: set_node_options
+  - name: dib-build
+    files: roles/nodepool/files/etc/nodepool/elements/*
 
 projects:
   - name: BonnyCI/bonnyci.org
@@ -92,8 +94,10 @@ projects:
   - name: BonnyCI/hoist
     check_github:
       - bonnyci-run-check-multinode
+      - dib-build
     gate_github:
       - bonnyci-run-gate-multinode
+      - dib-build
 
   - name: BonnyCI/sandbox
     check_github:


eggshell:
recheck

roles/zuul/files/jobs/dib.yaml
```diff
@@ -0,0 +1,38 @@
...
+    node: '{node}'
+    builders:
+      shell: |
+        virtualenv $HOME/dib-virtualenv
+        . $HOME/dib-virtualenv/bin/activate
+        pip install diskimage-builder
+
+        git clone https://github.com/BonnyCI/hoist.git $HOME
+        export ELEMENTS_PATH=$HOME/hoist/roles/nodepool/files/etc/nodepool/elements/nodepool
+        disk-image-create -a amd64 -o test-ubuntu vm ubuntu
```
gandelman-a: So we want to actually test an image build with all the elements we have in-repo.  The build command we actually end up running in production is:

disk-image-create -x -t qcow2 --checksum --no-tmpfs --qemu-img-options 'compat=0.10' -o /opt/nodepool/images/ubuntu-xenial-0000000037 ubuntu-minimal vm simple-init growroot openssh-server devuser haveged pip-and-virtualenv nodepool bonnyci-nodepool

Not sure if want to hard-code that into the job definition here,  parse the elements-to-test out of a 'nodepool_diskimages' variable somewhere in inventory, or just try to build an image with all elements listed in $HOME/hoist/roles/nodepool/files/etc/nodepool/elements/nodepool ?

roles/zuul/templates/etc/zuul/config/layout.yaml
```diff
@@ -92,8 +92,10 @@ projects:
   - name: BonnyCI/hoist
     check_github:
       - bonnyci-run-check-multinode
+      - dib-build-check
     gate_github:
       - bonnyci-run-gate-multinode
+      - dib-build-gate
 
   - name: BonnyCI/sandbox
     check_github:
```
gandelman-a: So we should be able to only run the dib-build-pipeline job when files in the roles/nodepool/files/etc/nodepool/elements are changed by the patch under test.  See https://docs.openstack.org/infra/zuul/zuul.html#jobs

I *think* we just need to add an entry to the layouts "jobs" section that applies the "files:" option to the dib job.

gandelman-a (commented):
Added some comments.  Would like to get input from others on the team about whether we should be adding an extra job to our pipeline, or to continue to use the vanilla bonnyci job to keep up good dog foodin' and avoid taking advantage of features that we dont provide our users

roles/zuul/files/jobs/dib.yaml
```diff
@@ -0,0 +1,38 @@
+- publisher:
```
ConsoleCatzirl: This publisher is already defined in bonnyci.yaml, I don't think we need to redefine it here.

roles/zuul/files/jobs/dib.yaml
```diff
@@ -0,0 +1,38 @@
...
+    publishers:
+      - scp:
+          site: bonnyci-scp
+          files:
+            - target: 'logs/$LOG_PATH'
+              source: 'logs/**'
+              keep-hierarchy: true
+              copy-after-failure: true
+
+- job-template:
```
ConsoleCatzirl: We're not doing anything different in the two pipelines, so I think we can just create a single job rather than use a job-template to build out multiple jobs.

ConsoleCatzirl (dismissed):

roles/zuul/files/jobs/dib.yaml
eggshell: @jesusaurus thanks for the feedback! I updated accordingly.

roles/zuul/templates/etc/zuul/config/layout.yaml
ConsoleCatzirl: I agree, we should limit this job to only run when the elements are changed.

eggshell: @gandelman-a @jesusaurus updated.

ConsoleCatzirl (commented):
+1 LGTM

roles/zuul/files/jobs/dib.yaml
```diff
@@ -0,0 +1,19 @@
...
+        virtualenv $HOME/dib-build
+        . $HOME/dib-build/bin/activate
+        pip install diskimage-builder
+
+        git clone https://github.com/BonnyCI/hoist.git /opt/hoist
+        export ELEMENTS_PATH=/opt/hoist/roles/nodepool/files/etc/nodepool/elements/nodepool
+        disk-image-create -x -t qcow2 --checksum --no-tmpfs     \
+                          --qemu-img-options 'compat=0.10'      \
+                          -o /opt/nodepool/images/ubuntu-xenial \
+                          ubuntu-minimal vm
```
gandelman-a: We need to specify the elements from the hoist repo that we want to include in the build and test.

gandelman-a (changes requested):

roles/zuul/files/jobs/dib.yaml
eggshell: @gandelman-a isn't that what `ELEMENTS_PATH` specifies above?

ConsoleCatzirl: @eggshell no, the elements path is where to find the listed elements, but this command only lists the two elements `ubuntu-minimal` and `vm`. I don't think there's a way to re-use the elements list from the `nodepool_diskimages` variable since it's scoped to a different host and all of these zuul jobs are files not templates, so for now the best thing might be to copy those elements into this command and add comments about keeping the two lists of elements in sync.

ConsoleCatzirl (changes requested):

roles/zuul/files/jobs/dib.yaml
```diff
@@ -0,0 +1,23 @@
+- job:
+    name: 'dib-build'
+    node: 'ubuntu-xenial'
+    builders:
+      shell: |
+        virtualenv $HOME/dib-build
+        . $HOME/dib-build/bin/activate
+        pip install diskimage-builder
+
+        git clone https://github.com/BonnyCI/hoist.git /opt/hoist
```
gandelman-a: The bonnyci user isn't going to have access to create this directory. Need to sudo this or create it in ~

roles/zuul/files/jobs/dib.yaml
```diff
@@ -0,0 +1,23 @@
...
+        export ELEMENTS_PATH=/opt/hoist/roles/nodepool/files/etc/nodepool/elements/nodepool
+        # NOTE: be sure the list of nodepool elements here stays consistent
+        # with the elements specified in the nodepool role.
+        disk-image-create -x -t qcow2 --checksum --no-tmpfs                 \
+                          --qemu-img-options 'compat=0.10'                  \
+                          -o /opt/nodepool/images/ubuntu-xenial             \
+                          ubuntu-minimal vm simple-init growroot            \
+                          openssh-server devuser haveged pip-and-virtualenv \
+                          nodepool bonnyci-nodepool
```
gandelman-a: I'd love to see if we can extract the list of elements we want to test from hoist itself, if its possible.  This static list is fine for testing what we currently ship, but if we add a new element to hoist, we have no way of testing it pre-merge until the job here is updated.

gandelman-a (changes requested):
Couple more comments.

Also, we need the job to install some distro package requirements for dib builds to work on the slave:
https://github.com/BonnyCI/hoist/blob/master/roles/nodepool/tasks/main.yml#L4

mlangbehn (commented):
This looks good to me.

roles/zuul/files/jobs/dib.yaml
```diff
@@ -0,0 +1,26 @@
...
+        sudo apt-get install -y debootstrap qemu-utils curl \
+                                uuid-runtime parted kpartx
+
+        sudo git clone https://github.com/BonnyCI/hoist.git /opt/hoist
+        export ELEMENTS_PATH=/opt/hoist/roles/nodepool/files/etc/nodepool/elements/nodepool
+        # NOTE: be sure the list of nodepool elements here stays consistent
+        # with the elements specified in the nodepool role.
+        disk-image-create -x -t qcow2 --checksum --no-tmpfs                 \
+                          --qemu-img-options 'compat=0.10'                  \
+                          -o /opt/nodepool/images/ubuntu-xenial             \
```
gandelman-a: It looks like there'll also be an issue with the bonnyci user writing the image to /opt/nodepool/images/ubuntu-xenial

gandelman-a (changes requested):

roles/zuul/files/jobs/dib.yaml
```diff
@@ -0,0 +1,26 @@
...
+        pip install diskimage-builder
+
+        sudo apt-get install -y debootstrap qemu-utils curl \
+                                uuid-runtime parted kpartx
+
+        sudo git clone https://github.com/BonnyCI/hoist.git /opt/hoist
+        export ELEMENTS_PATH=/opt/hoist/roles/nodepool/files/etc/nodepool/elements/nodepool
+        # NOTE: be sure the list of nodepool elements here stays consistent
+        # with the elements specified in the nodepool role.
+        sudo disk-image-create -x -t qcow2 --checksum --no-tmpfs            \
```
gandelman-a, line 17: You loose the virtualenv environment when you sudo here, so disk-image-create is not in $PATH.  Might make more sense to just write the image to /tmp or somewhere in $HOME instead?

roles/zuul/files/jobs/dib.yaml
```diff
@@ -0,0 +1,26 @@
...
+    builders:
+      shell: |
+        virtualenv $HOME/dib-build
+        source $HOME/dib-build/bin/activate
+        pip install diskimage-builder
+
+        sudo apt-get install -y debootstrap qemu-utils curl \
+                                uuid-runtime parted kpartx
+
+        sudo git clone https://github.com/BonnyCI/hoist.git /opt/hoist
```
gandelman-a, line 13: So cloning and using the upstream repository means we wouldn't actually be testing the proposed changes here, and instead using whats already been merged.  When the test runs, it'll be running from the root of the hoist repo with the proposed PR already checked out, so you should be able to just point ELEMENTS_PATH to ./roles/nodepoole/files/etc/nodepool/elements/

gandelman-a (changes requested):
A few more things...  FWIW I'm able to test this locally in a fresh ubuntu vm by just copying the shell snippet into a file and running it there.  With my comments addressed, the build still quietly fails in elements/nodepool/extra.d/60-copy-nodepool-scripts with a pushd error.

PR #357, opened 2017-04-20
Add pruner script to ARA role

rattboi:
ARA currently is saving entries to a sqlite db on the bastion. We run
ansible *a lot* so this database is getting quite large. This script
uses ARA's cli tool to determine ara playbooks older than a certain
date, and deletes them from the database.

This is a stop-gap solution until I move the ARA database to mysql on
the logs host.

### DIFF
--- a/roles/ara/defaults/main.yml
+++ b/roles/ara/defaults/main.yml
@@ -1,0 +1,4 @@
+# Keep 7 days worth of ARA data
+ara_prune_days: 7
+#Run offset from ansible-runner :07,:22,:37,:52
+ara_prune_cron_schedule: 7,22,37,52
--- a/roles/ara/files/usr/local/bin/ara-prune
+++ b/roles/ara/files/usr/local/bin/ara-prune
@@ -1,0 +1,32 @@
+#!/bin/bash -eu
+
+if [[ -z ${1+x} ]]; then
+  echo "ERROR: Must supply argument for # of days to keep"
+  exit 1
+else
+  days_to_keep="$1"
+fi
+
+ara_venv=${ANSIBLE_RUNNER_VENV:-/opt/ansible}
+
+set +u
+. "$ara_venv"/bin/activate
+set -u
+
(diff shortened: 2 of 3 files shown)


roles/ara/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+# Keep 7 days worth of ARA data
+ara_prune_days: 7
+#Run offset from ansible-runner :07,:22,:37,:52
+ara_prune_cron_schedule: 7-59/15
```
ConsoleCatzirl: This won't actually do what you expect, it will be parsed as (7-59)/15, which ends up being :15, :30, :45

From `man 5 crontab`:
```
Step  values  can  be  used in conjunction with ranges.  Following a range with ``/<number>'' specifies skips of
the number's value through the range.  For example, ``0-23/2'' can be used in the hours field to specify command
execution every other hour (the alternative in the V7 standard is ``0,2,4,6,8,10,12,14,16,18,20,22''). Steps are also
permitted after an asterisk, so if you want to say ``every two hours'', just use ``*/2''.
```

Let's just explicitly list the four times with: `7,22,37,52`

roles/ara/files/usr/local/bin/ara-prune
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -eu
+
+days_to_keep="$1"
```
ConsoleCatzirl: Let's add some error checking to make sure that we have a non-empty value here.

ConsoleCatzirl (dismissed):

roles/ara/defaults/main.yml
rattboi: Fixed

roles/ara/files/usr/local/bin/ara-prune
rattboi: Fixed as well

eggshell (approved):

PR #358, opened 2017-04-21
Enable SSL on the bastion apache

jamielennox:
We're serving ARA, cron logs and other things on apache on the bastion.
As with all our other servers this should be done over SSL.

Again there's no real way to test this except on prod so only approve
when someone is there to shepherd it through.

### DIFF
--- a/bastion.yml
+++ b/bastion.yml
@@ -14,6 +14,10 @@
       tags:
         - monitoring
       when: secrets is defined
+    - role: letsencrypt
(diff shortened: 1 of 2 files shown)


jamielennox:
recheck

jamielennox:
recheck

ConsoleCatzirl:
LGTM

gandelman-a (approved):

PR #359, opened 2017-04-21
Add logging-proxy deployment to logs server

jamielennox:
At the moment we deploy the logging proxy onto the log server on a high
port. We don't need to expose this on a standard port for now because
people will only follow links. In future the client side of this can be
integrated into the bonnyci.org site and the websock be on whatever port
we like. For now splitting that up is not worth it.

I've set up the proxy so that it redirects through apache so that when
we put SSL on the site this will also be affected.

There's some logging magic in here to get from journald -> syslog ->
file so we can later put it into logstash.

### DIFF
--- a/install-ci.yml
+++ b/install-ci.yml
@@ -251,6 +251,20 @@
             socket: localhost:3310
             lazy-apps: true
 
+    - role: logging-proxy
+      logging_proxy_host: 127.0.0.1
+
+    - role: logrotate
+      logrotate_configs:
+        - name: zuul
+          path: /var/log/logging-proxy/*log
+          options:
+            - compress
+            - missingok
+            - rotate 30
+            - daily
+            - notifempty
+
     - role: apache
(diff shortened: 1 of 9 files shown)


jamielennox:
-1 Other than the failure i just made BonnyCI/logging-proxy#1 which will make the logging much easier.

install-ci.yml
```diff
@@ -251,6 +251,20 @@
             socket: localhost:3310
             lazy-apps: true
 
+    - role: logging-proxy
+      logging_proxy_host: 127.0.0.1
+
+    - role: logrotate
+      logrotate_configs:
+        - name: zuul
```
ConsoleCatzirl, line 259: This name should probably be logging-proxy. I think this might overwrite the zuul config if both are included on the same host.

ConsoleCatzirl (changes requested):
For the most part this looks good

jamielennox:
For an alternative: https://github.com/BonnyCI/hoist/pull/361

PR #361, opened 2017-04-24
Add a docker and logging-proxy role

jamielennox:
Instead of installing nodejs and everything in place go full devops and
docker it. Quay.io can auto build containers on every github push and
then ansible just ensures that the latest container from the repository
is installed and running.

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -1,0 +1,41 @@
+# The ansible version of https://docs.docker.com/engine/installation/linux/ubuntu/
+---
+- name: Remove older package names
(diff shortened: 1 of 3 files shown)


omgjlk (commented):
I think this is neat, but may also require some lore around it for how to manage it.

I'm a +1 on this, but I'll let somebody else approve.

jamielennox:
for now we are going to not expose the streaming logs and focus on doing it right in v3 - but yea, i think the approach is better and something worth exploring for everythigng.

ConsoleCatzirl (approved):

PR #363, opened 2017-04-25
ansible-runner: Drop flag files for failing environments

gandelman-a:
This updates ansible-runner to drop empty files in a known
directory for environments that are failing to complete their
playbooks. The files are named after the failing environment.
Upon successful ansible run, these files are cleaned up if
they exist.

This'll allow us to set a datadog check that fails when any
files exist here.  Our current datadog check is broken and
relies on log scraping and doesn't really work with multi-env
bastions.

Related-Issue: BonnyCI/projman#224


ConsoleCatzirl (approved):
Yay, more tests!

PR #364, opened 2017-04-25
Adds dd-ansible-runner check and role

gandelman-a:
This adds a simple datadog check that fails if it finds any flag files for
failing ansible-runner environments.  This depends on PR BonnyCI/hoist#363
but should pass OK if it lands before that merges.

Related-Issue: BonnyCI/projman#224

### DIFF
--- a/bastion.yml
+++ b/bastion.yml
@@ -36,6 +36,9 @@
             - notifempty
       when: secrets is defined
 
+    - role: dd-ansible-runner
+      tags:
+        - monitoring
     - role: ansible-runner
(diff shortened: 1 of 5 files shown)


jamielennox:
-1

So there is already a callback for ansible and datadog - i know i get an email every time it fails which is pretty annoying. I think the information is there we just need to fix the alert to something more useful.

jamielennox:
see: https://app.datadoghq.com/monitors#1617580?group=all&live=4h

PR #365, opened 2017-04-25
Use explicit bastion ssl flag instead of letsencrypt

jamielennox:
This bit us in zuul as well. We should be able to use the presence of
the letsencrypt variable as an indicator to turn on SSL but the existing
string evaluates as False. Make it explicit. This is annoying because
you need to set 2 flags to turn on SSL.


omgjlk (approved):

gandelman-a:
gate-recheck

PR #366, opened 2017-04-26
Set datadog hostname on elk, logs and bastion

jamielennox:
It looks like datadog is considering bastion and bastion.openstack and
the others as seperate hosts because what is in the ansible inventory
does not match what datadog thinks it is talking to.

Hopefully if we force set the hostname these things will start being
correlated. And I'm not sure if they're being billed as seperate
entities or not.


jamielennox:
Prefer: https://github.com/BonnyCI/hoist/pull/367

PR #370, opened 2017-04-26
Set datadog hostname to inventory_hostname

jamielennox:
*sigh* attempt 3 at this. The whole point is to get the names reported
in datadog to match up to those that ansible reports. To do that we need
to match them to things in the inventory.

That requires using inventory_hostname, not ansible_hostname which will
just set them to the same discovered hostname that datadog would anyway.

### DIFF
--- a/bastion.yml
+++ b/bastion.yml
@@ -12,7 +12,7 @@
     - common
 
     - role: dd-agent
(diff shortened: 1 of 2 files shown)


jamielennox:
Nope. Fix in properly in #371.

PR #371, opened 2017-04-26
Standardize hostnames on production servers

jamielennox:
We are running into the situation where our monitoring is getting confused as
to which host emitted an event and mismatching ansible and host output because
the recognized hostname is different.

Standardize all our production hostnames.

Fixes: BonnyCI/projman#230


bastion.yml
```diff
@@ -9,10 +9,10 @@
         mode: 0700
 
   roles:
-    - common
+    - role: common
+      common_hostname: "{{ bonnyci_hostname | default('') }}"
```
gandelman-a: I dont see bonnyci_hostname being set in any of these commits. Will that end up being set to match ansible_host or inventory hostname?

gandelman-a (commented):
I'm okay with this. It'll require some DNS work when we role it out, will wait to approve it until some others have taken a look and we can coordinate a plan for approving it through and deploying it.

jamielennox:
@gandelman-a as it stands there are no DNS changes required, all i've done is change the host's understanding of its own name - and that's in /etc/hosts anyway. 

I think standardizing the DNS might not be a bad idea either but it isn't required for this.

omgjlk (commented):
This all seems reasonable. I'd approve, if we're ready to see it deployed.

ConsoleCatzirl (commented):
LGTM

gandelman-a (approved):

loafyloaf (commented):
awesome

PR #372, opened 2017-04-26
WIP: Add ansible runner logs to ELK

### LINKED ISSUE #28: Figure out system log publishing
_From @missaugustina on November 23, 2016 21:38_

where are the logs for our stuff that's running?

- [ ] system logs (ansible crons)
- [x] daemon logs (nodepool, zuul)
- [x] drop debug logs (too noisy)
- [x] Spike on aggregating logs on a single host
- [x] Consider an ELK stack in addition
- [ ] Tune log parsing for our service logs

_Copied from original issue: BonnyCI/hoist#99_

loafyloaf:
fixes: BonnyCI/projman#28

### DIFF
--- a/bastion.yml
+++ b/bastion.yml
@@ -114,9 +114,12 @@
         - name: ansible_runner
           prospectors:
             - input_type: log
-              document_type: ansible_runner
+              document_type: ansible-runner
               paths:
-                - "/var/www/html/cron-logs/*/current.log"
-                - "/var/www/html/cron-logs/*/latest.log"
+                - "/var/www/html/cron-logs/*/*current.log"
+              multiline:
+                pattern: '^$'
+                negate: true
+                match: before
 
     - role: fail2ban  # This should be last
--- a/bastion.yml
+++ b/bastion.yml
@@ -111,5 +111,11 @@
               document_type: apache
               paths:
                 - /var/log/apache/*.log
+        - name: ansible_runner
+          prospectors:
+            - input_type: log
+              document_type: ansible_runner
+              paths:
+                - "/var/www/html/cron-logs/$env/*.log"
 
     - role: fail2ban  # This should be last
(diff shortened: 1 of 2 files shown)


bastion.yml
```diff
@@ -111,5 +111,11 @@
               document_type: apache
               paths:
                 - /var/log/apache/*.log
+        - name: ansible_runner
+          prospectors:
+            - input_type: log
+              document_type: ansible_runner
+              paths:
+                - "/var/www/html/cron-logs/$env/*.log"
```
ConsoleCatzirl: We should only track either the current or the latest log, but if we track all logs in the directory then we will get a bunch of duplicate entries after every log rotation. Also there is nothing setting `$env`, so that should be another `*` instead.

Also, the document type is used by logstash to know how to parse the logs. Using `ansible_runner` for the type is correct, but doesn't have a matching logstash filter yet, we will need to add that as well. We will need a change similar to https://github.com/BonnyCI/hoist/commit/808c70cd50981dc9f5eaef560aacf9af02faacef to add the new log parsing. And we can probably reuse the exact same multiline configuration from that change since it is based on lines starting with a timestamp. The tricky part is the grok pattern, we can chat about that in irc or slack.

ConsoleCatzirl (changes requested):
A good start, but we need a little more than this.

roles/logstash/files/etc/logstash/conf.d/16-ansible-runner.conf
```diff
@@ -0,0 +1,10 @@
+filter {
+  if [type] == "ansible-runner" {
+      grok {
+        match => { "message" => "%{SUPERAWESOMEANSIBLELOG}"}
```
ConsoleCatzirl: I'll take a stab at writing our super-awesome ansible-log pattern

PR #376, opened 2017-04-27
Remove portbleu host_vars

jamielennox:
We are no longer running the bonnyci-internal.portbleu.com environment
so we can remove these host vars.


loafyloaf (approved):
makes sense to me

PR #377, opened 2017-04-27
Fix ara-pruner to find where the ARA db is

rattboi:
The ARA cli program looks for ansible.cfg in a few places, and if it
doesn't find it, it uses some defaults and creates a new ARA db.

This leads the ara-pruner script to not be looking at the right
database. This points it at the right place.


rattboi:
I'm abandoning this for now, because activating it now will probably slow the bastion WAY down for a while.

PR #378, opened 2017-04-27
Update vagrant nodepool provider

ConsoleCatzirl:
Update the nodepool provider defined in the vagrant group vars to
work with the current opentechsjc network setup. There are multiple
networks, so we must specify one to use. Also reduce the minimum nodes
to reduce resource overhead.

### DIFF
--- a/inventory/group_vars/vagrant
+++ b/inventory/group_vars/vagrant
@@ -27,12 +27,15 @@ nodepool_labels:
 nodepool_providers:
   - name: cicloud
     cloud: cicloud
-    max-servers: 10
+    max-servers: 1
     images:
       - name: ubuntu-xenial
-        min-ram: 2048
+        min-ram: 512
         diskimage: ubuntu-xenial
         private-key: /var/lib/nodepool/.ssh/id_rsa
+    networks:
+      - name: 'internal'
+        public: False
 nodepool_mysql_host: zuul.vagrant
 nodepool_gearman_servers:
   - host: zuul.vagrant


inventory/group_vars/vagrant
```diff
@@ -27,12 +27,15 @@ nodepool_labels:
 nodepool_providers:
   - name: cicloud
     cloud: cicloud
-    max-servers: 10
+    max-servers: 3
```
gandelman-a: for developers, we should probably keep this to 1 by default just to ease resource usage on the cloud

inventory/group_vars/vagrant
```diff
@@ -27,12 +27,15 @@ nodepool_labels:
...
     images:
       - name: ubuntu-xenial
         min-ram: 2048
```
gandelman-a: we can bump this down to 512 as well to conserve resources

inventory/group_vars/vagrant
```diff
@@ -27,12 +27,15 @@ nodepool_labels:
...
         diskimage: ubuntu-xenial
         private-key: /var/lib/nodepool/.ssh/id_rsa
+    networks:
+      - name: 'nodepool'
+        public: False
```
gandelman-a: I dont think we want to encourage devs to boot slaves on our production nodepool environment.  We could instead default to the 'internal' network and the slaves should be reachable via fixed ip from the dev env, assuming it to was booted on the internal network.

gandelman-a (dismissed):
I have a similar change to the allinone inventory i was just about to push up.... a couple comments

inventory/group_vars/vagrant
gandelman-a: Also, tip: we also need user-home and username set on the image so that nodepool, otherwise it defaults to jenkins

PR #379, opened 2017-04-28
Remove the default_deny on monitoring vhost

jamielennox:
The default deny rule on the monitoring host is conflicting with the
letsencrypt role that is also listening on this port. It's kind of hard
to get the servername through to letsencrypt to deconflict them so just
remove the deny rule.

### DIFF
--- a/inventory/group_vars/monitoring
+++ b/inventory/group_vars/monitoring
@@ -5,12 +5,7 @@ bonnyci_kibana_apache_mods_enabled:
 
 bonnyci_kibana_apache_vhosts:
   - name: default_deny
-    server_name: default.only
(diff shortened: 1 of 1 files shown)


jamielennox:
recheck

omgjlk (commented):
LGTM, I'll let somebody else approve.

gandelman-a (commented):
Hmm we added this rule because a proxying service picked up this host and was sending TONS of requests to :80.  It looks like deny'ing those requests has since removed the host so this is probably ok. Will let someone else +a

rattboi (approved):
LGTM. We should circle back once @jesusaurus is available.

PR #380, opened 2017-05-01
Use a branch for the zuul v3 environment

jamielennox:
For the next little while until the github patches are all merged
upstream redirect our zuul v3 efforts onto our own github branch. This
will let us experiment and test more effectively.

### DIFF
--- a/inventory/group_vars/opentech-sjc-v3
+++ b/inventory/group_vars/opentech-sjc-v3
@@ -6,8 +6,8 @@ zuul_zuul_v3: True
 nodepool_git_repo_url: https://github.com/openstack-infra/nodepool
 nodepool_git_version: feature/zuulv3
 
-zuul_git_repo_url: https://github.com/openstack-infra/zuul
-zuul_git_branch: feature/zuulv3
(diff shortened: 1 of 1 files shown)


tests/files/validate-ci.yml
```diff
@@ -17,10 +17,6 @@
 - name: Validate zuul status
   hosts: zuul
   tasks:
-    - name: zuul webapp is running
-      uri:
-        url: http://localhost:8001/status.json
```
omgjlk: Was this taken out on purpose?

omgjlk (commented):
I think something extra got in here.

tests/files/validate-ci.yml
jamielennox: yea, the new url is /{tenant_name}/status.json - but we don't have a default tenant we can assert the test with. Was trying to think of some way of doing it conditionally, or assume a tenant, and then just removed it.

gandelman-a: validate-ci.yml doesnt ever run against a v3, so we should be able to avoid dropping this

gandelman-a (approved):

PR #381, opened 2017-05-01
Copy zuul and nodepool logs into bonny reporter

jamielennox:
We often want to see what went wrong with a hoist job, copy the zuul and
nodepool logs for now as these are the most likely to fail.


omgjlk (approved):
Looks like it'll do the right thing.

PR #384, opened 2017-05-03
Convert ARA role into separate roles for each piece

rattboi:
Fixes BonnyCI/projman#173

Split into 3 parts: ansible callback, db, web-ui
 - ARA callback is an optional piece of ansible-runner
 - ARA db is mysql-backed
 - ARA web-ui is a flask app, and needs separate pieces for apache + uwsgi

I also removed ara-prune for now, but if we need it, we can add it back
later.


roles/ansible-runner/tasks/main.yml
```diff
@@ -83,3 +70,6 @@
 
 - include: runner.yml
   with_items: "{{ ansible_runner_tasks }}"
+
+- when: ara_enabled | default(False) | bool
+  include: ara.yml
```
omgjlk: Can you flip these? Do the include, with the conditional after? Makes more sense as the conditional is applied to every task within, rather than making the inclusion itself conditional.

omgjlk (changes requested):
Minor nit on the formatting, otherwise this looks fine!

omgjlk (approved):

PR #385, opened 2017-05-03
Redirect all zuul web requests to the app

jamielennox:
Zuul V3 has tenant names in the URLs so they are much more variable.
This means that we can't simply rely on hardcoding a couple of paths
into our apache config and we should probably just forward everything to
the zuul app.

### DIFF
--- a/inventory/group_vars/zuul
+++ b/inventory/group_vars/zuul
@@ -1,7 +1,6 @@
 bonnyci_zuul_webapp_apache_mods_enabled:
   - proxy.load
   - proxy_http.load
-  - rewrite.load
 
 bonnyci_zuul_webapp_apache_vhosts:
   - name: status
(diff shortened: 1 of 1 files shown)


ConsoleCatzirl:
Doesn't this mean that we will no longer be serving /opt/source/zuul/etc/status/public_html/ and so no longer serving the website displaying the states of the pipeline queues? If so, should we create a new vhost to serve that public_html directory so that we can still have a website showing the states of the pipeline queues?

omgjlk:
I think the web app itself takes care of the pipeline states.

jamielennox:
I... don't know. The webapp is broken on v3 so i haven't been able to check.

Is there a way we can prioritize serving the static files?

ConsoleCatzirl:
It looks like rewrite rules allow for PCRE, so we can glob the tenant portion similar to how we glob the current status rewrite.

PR #390, opened 2017-05-04
Move ara-db stuff into opentech-sjc-common directory

gandelman-a:
Both the bastion and the logs host are both deployed as part of
the opentech-sjc-common inventory.  This moves the required
ARA stuff into there and out of the zuul-specific inventories,
which dont deploy either.


rattboi (approved):
:+1: Sorry I missed this.

PR #403, opened 2017-05-09
Give each ansible-runner task its own virtualenv

gandelman-a:
We currently share a base virtualenv among ansible-runner tasks, into
which we install ansible, datadog callbacks, ara and requirements for
all ansible-runners.  This is problematic because 1) ansible-runners
that do not run as root require sudo access to update the virtualenv
2) we can potentially end up with version conflicts among dependencies

Instead, this sets up isolated virtualenvs for each runner task, each with
their own callbacks, dependency set, etc.

Closes-issue: BonnyCI/projman#242

### DIFF
--- a/roles/ansible-runner/tasks/ara.yml
+++ b/roles/ansible-runner/tasks/ara.yml
@@ -2,9 +2,9 @@
 - name: Install ara
   pip:
     name: "ara>=0.13.0"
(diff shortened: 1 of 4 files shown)


ConsoleCatzirl (commented):
+1 from me, but I think this could use a second set of eyes before approving.

omgjlk:
recheck

PR #406, opened 2017-05-10
Put the correct integration key in place in v3

jamielennox:
We have some level of abstraction in specifying the file for the
integration key in v2 and v3, however in both environments we were
actually still copying the v2 key. This means that v3 can't authenticate
correctly.

Fix that to put the v3 key in place in the v3 environment.


rattboi (approved):
LGTM :+1:

PR #409, opened 2017-05-19
Remove bot host as it is no longer needed

SpamapS:
Sadly, we don't need quartermaster at the moment. We can keep the
playbook and roles around in case we ever want to turn it back on again.

Signed-off-by: Clint Byrum <clint@fewbar.com>


omgjlk (approved):
/me pours one out for quartermaster.

PR #411, opened 2017-05-23
Nginx

jamielennox:
Add nginx support.

We'll need to stop the zuul v3 job, merge this, and restart the job. I don't have any real way to test the letsencrypt work so that will likely require fixing.


roles/letsencrypt/tasks/main.yml
```diff
@@ -125,7 +125,7 @@
     data: "{{ letsencrypt_challenge_response }}"
     dest: "{{ letsencrypt_cert_path }}"
     remaining_days: "{{ letsencrypt_remaining_days }}"
-  notify: restart apache
+  notify: "{{ letsencrypt_reload_handler }}"
```
omgjlk, line 128: We could probably solve this by doing a generic notify topic of "restart web server" or something like that, and have both the apache and nginx handlers listen to that topic. But maybe that'll be weird if we wind up with both on the same host?

install-ci.yml
```diff
@@ -14,11 +14,19 @@
       tags:
         - monitoring
 
-    - role: letsencrypt
+    - role: letsencrypt-apache
       tags:
         - letsencrypt
       when:
         - letsencrypt_csr_cn | default(False)
+        - bonnyci_use_apache
```
omgjlk, line 22: I'm a little concerned that these conditional items aren't piped through the bool filter, but not enough to block this.

omgjlk (approved):
Hold on to your butts!

PR #413, opened 2017-05-24
Add a combined certificate for letsencrypt

jamielennox:
Nginx requires that we combined all the certificates into one combined
file. Add this functionality to the letsencrypt role.

### DIFF
--- a/roles/letsencrypt/tasks/main.yml
+++ b/roles/letsencrypt/tasks/main.yml
@@ -17,6 +17,7 @@
     - "{{ letsencrypt_key_path | dirname }}"
     - "{{ letsencrypt_cert_path | dirname }}"
     - "{{ letsencrypt_chain_path | dirname }}"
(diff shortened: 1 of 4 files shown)


roles/letsencrypt/tasks/main.yml
```diff
@@ -128,6 +129,25 @@
   notify: "{{ letsencrypt_reload_handler }}"
   when: letsencrypt_challenge_response.changed
 
+- name: Read certificates
+  command: "cat {{ item }}"
+  register: letsencrypt_certificates
+  changed_when: False
+  with_items:
+    - "{{ letsencrypt_cert_path }}"
+    - "{{ letsencrypt_chain_path }}"
```
omgjlk, line 138: Instead of cat and register, you could use the `slurp` module to read the file into a variable.

omgjlk (approved):
slurp would be more efficient, but cat is fine. We can always tweak it later.

PR #417, opened 2017-05-25
Add integration-handler setup

jamielennox:
The integration-handler will handle webhooks specific to how we actually
setup our integration. Things like when a new installation was created
and we want to add that to our tenant file. Nginx will look at the event
type header and redirect certain events off to the handler.

### DIFF
--- a/roles/integration-handler/tasks/main.yml
+++ b/roles/integration-handler/tasks/main.yml
@@ -1,0 +1,22 @@
+---
+- name: create invoke script
+  template:
+    src: invoke.sh.j2
+    dest: "{{ integration_handler_invoke_path }}"
+    mode: 0755
+    owner: root
+
+- name: create sudoers file
+  template:
+    src: etc/sudoers.d/90-bonnyci-integration-handler
+    dest: /etc/sudoers.d/90-bonnyci-integration-handler
+    mode: 0440
+    owner: root
+    validate: 'visudo -cf %s'
+
+- name: Trigger run if tenant file doesn't exist
+  stat:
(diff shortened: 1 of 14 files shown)


roles/integration-handler/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+    mode: 0440
+    owner: root
+    validate: 'visudo -cf %s'
+
+- name: Trigger run if tenant file doesn't exist
+  stat:
+    path: "{{ integration_handler_zuul_tenant_file }}"
+  register: integration_handler_tenant_stat
+  changed_when: not integration_handler_tenant_stat.stat.exists
+  notify: Run integration-handler
```
omgjlk, line 22: Looks like this is the only time we notify that handler. COuldn't we simply just do a `command` task that has a `creates` argument pointing to the file path? That way if the file exists the task is considered done, but if it doesn't exist it'll run the command.

Not worth blocking the review, just pointing out an optimization.

omgjlk (approved):
I think this is okay. Hard to tell what will happen when the file doesn't exist. Because that code lives somewhere else. But looks right.

roles/integration-handler/tasks/main.yml
jamielennox: Yea, probably. There was one other case where i thought you would want to do a manual run which is when the template file is updated over in zuul. But i couldn't see (didn't try that hard) a clean way to make the handler work across both. 

Real encapsulation in ansible is a PITA.

PR #430, opened 2017-07-03
Rename github integration to app

jamielennox:
With the public release of the integration github changed the name to
app. In turn we updated the github driver in zuul to use the word app
instead. Fix hoist to use these terms.


omgjlk (approved):
Did the secrets file change too?

PR #437, opened 2017-08-09
Add sshd role to lockdown ssh

jamielennox:
With the change to softlayer we end up having ssh root login enabled by
default. Create an sshd role that will lock down SSH root access and
some other security tweaks.

### DIFF
--- a/roles/sshd/tasks/main.yml
+++ b/roles/sshd/tasks/main.yml
@@ -1,0 +1,20 @@
+- name: Install SSHD
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - openssh-server
+
+- name: setup sshd config
+  template:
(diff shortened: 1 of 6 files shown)


roles/sshd/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+    - openssh-server
+
+- name: setup sshd config
+  template:
+    src: etc/ssh/sshd_config
+    dest: /etc/ssh/sshd_config
+    owner: root
+    mode: "0644"
+  notify: Restart sshd
+
```
omgjlk, line 15: Probably should have done a flush handlers here, otherwise there COULD be a start in the next task, followed by a restart once the role finishes. That said, we use ssh to do these tasks, so it'd be a pretty weird scenario to not have SSH running for the below task :)

PR #486, opened 2017-08-21
Split zuul into components

jamielennox:
We're not going to be installing so many components side by side in
future and we need to seperate which pieces belong to which zuul.

This is going to be a big commit that'll likely require fixing.


roles/zuul-executor/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+    state: present
+    update_cache: yes
+
+- name: Install bubblewrap
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - bubblewrap
+
```
omgjlk, line 53: Probably should have a handler flush here, so that you don't have a start followed by a restart, but this is a minor nit.

roles/zuul-scheduler/tasks/main.yml
```diff
@@ -48,3 +52,8 @@
   get_url:
     url: https://maxcdn.bootstrapcdn.com/bootstrap/3.1.1/css/bootstrap.min.css
     dest: /opt/source/zuul/etc/status/public_html/lib/bootstrap/css/bootstrap.min.css
+
```
omgjlk, line 55: and a flush here too.

roles/zuul-server/tasks/main.yml
```diff
@@ -45,3 +57,9 @@
   get_url:
     url: https://maxcdn.bootstrapcdn.com/bootstrap/3.1.1/css/bootstrap.min.css
     dest: /opt/source/zuul/etc/status/public_html/lib/bootstrap/css/bootstrap.min.css
+
```
omgjlk, line 60: and flush here.

omgjlk (dismissed):
A few minor nits, but those can be fixed in follow up PRs.

roles/zuul-executor/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+    host: "{{ item.host }}"
+  with_items: "{{ zuul_ssh_known_hosts }}"
+  when: zuul_ssh_known_hosts is defined
+
+- name: Ensure zuul ssh known_hosts ownership
+  file:
+    path: "{{ zuul_home_dir }}/.ssh/known_hosts"
+    owner: zuul
+    group: zuul
+  when: zuul_ssh_known_hosts is defined
```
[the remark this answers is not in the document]
jamielennox: So this is copied over - but i'm not sure. Thre doesn't seem to be an owner/group option to the known_hosts module and we are executing as root and setting to zuul so i can see it could fail. But in which case it would be really good if ansible just fixed this up for us.

omgjlk (approved):

PR #511, opened 2017-09-21
Add CentOS 7 support to hoist

SpamapS:


roles/common/tasks/main.yml
```diff
@@ -3,12 +3,23 @@
...
     state: installed
+  when: ansible_os_family == "Debian"
   with_items:
     - build-essential
     - jq
     - netcat
     - net-tools
 
+- name: Install common packages
+  yum:
```
jamielennox: when: ansible_os_family == "Redhat" ??

roles/apache/tasks/main.yml
```diff
@@ -1,4 +1,8 @@
 ---
 - include: apache.yml
+  when: ansible_os_family == "Debian"
```
jamielennox: nit: this seems reasonable, but if we're going to do apache_redhat can we rename this apache_ubuntu

roles/apache/tasks/ssl_redhat.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: set stapling files
+  copy:
+    src: stapling.conf
+    dest: /etc/httpd/conf.d/stapling.conf
+    mode: 0644
+    owner: root
+    group: root
+  notify: restart apache
```
jamielennox, line 9: feels like this could be merged into existing with some vars, but i'm not going to get picky with all this.

roles/common/tasks/main.yml
jamielennox: the other way to do all this in include_vars: "{{ ansible_os_family }" and use package: instead of apt/yum. Whatever.

roles/sshd/tasks/main.yml
```diff
@@ -8,6 +8,6 @@
 
 - name: ensure ssh started
   service:
-    name: ssh
+    name: "{{ 'ssh' if ansible_os_family == 'Debian' else 'sshd' }}"
```
jamielennox, line 11: nit: not a blocker, but yea i would prefer to do include_var stuff rather than ifs inline

roles/apache/tasks/apache_redhat.yml
```diff
@@ -13,6 +13,12 @@
...
 
+- name: Disable default Listen
+  lineinfile:
+    regexp: ^Listen
+    line: "# Listen disabled"
+    path: /etc/httpd/conf/httpd.conf
+
 - meta: flush_handlers
 
 - name: start apache
```
jamielennox: squash into first commit?

roles/mysql/tasks/debian.yml
```diff
@@ -0,0 +1,42 @@
...
+     - mysql-server
+     - mysql-client
+     - python-mysqldb
+     - libmysqlclient-dev
+
+ - name: Set mysql_flavor
+   set_fact: mysql_flavor=mysql
+
+ - name: Set mysql_daemon_name
+   set_fact: mysql_daemon_name=mysqld
```
jamielennox, line 26: vars instead of set_fact?

roles/zookeeper/tasks/upstream_install.yml
```diff
@@ -0,0 +1,34 @@
...
+- name: Download Upstream ZK Tarball
+  get_url:
+    dest: /usr/local/zk/zookeeper-3.4.10.tar.gz
+    url: https://apache.org/dist/zookeeper/stable/zookeeper-3.4.10.tar.gz 
+
+- name: Untar Zookeeper
+  unarchive:
+    src: /usr/local/zk/zookeeper-3.4.10.tar.gz
+    dest: /usr/local
+    remote_src: yes
```
jamielennox, line 16: is this idempotent? personal opinion, i'd try and find an rpm.

roles/pip/vars/main.yml
```diff
@@ -1,2 +1,3 @@
 ---
 pip_python_exec: "{{ pip_python3 | bool | ternary('python3', 'python') }}"
+pip_exec: "{{ pip_python3 | bool | ternary('pip3', 'pip') }}"
```
jamielennox: ok, surprised that this isn't handled by the python that you installed it wiht.

roles/sshd/handlers/main.yml
```diff
@@ -1,4 +1,4 @@
 - name: Restart sshd
   service:
-    name: ssh
+    name: "{{ 'ssh' if ansible_os_family == 'Debian' else 'sshd' }}"
     state: restarted
```
jamielennox, line 4: squash,

roles/python-app/vars/main.yml
```diff
@@ -1,2 +1,3 @@
 ---
 python_app_python_exec: "{{ python_app_python3 | bool | ternary('python3', 'python') }}"
+python_app_pip_exec: "{{ pip_python3 | bool | ternary('pip3', 'pip') }}"
```
jamielennox: squash?

roles/nginx/tasks/redhat.yml
```diff
@@ -0,0 +1,40 @@
+- name: Install nginx
+  yum:
+    name:
+      - nginx
```
jamielennox, line 4: nit: this looks off. with_items?

roles/python-app/vars/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
-python_app_python_exec: "{{ python_app_python3 | bool | ternary('python3', 'python') }}"
-python_app_pip_exec: "{{ pip_python3 | bool | ternary('pip3', 'pip') }}"
+python_app_python_exec: "{{ python_app_python3 | bool | ternary('python3.5', 'python') }}"
+python_app_pip_exec: "{{ pip_python3 | bool | ternary('pip3.5', 'pip') }}"
```
jamielennox: feels like this would cause us upgrade problems.

roles/zuul-executor/tasks/main.yml
```diff
@@ -30,18 +30,26 @@
...
     name: "{{ item }}"
     state: present
   with_items:
     - bubblewrap
 
+- name: Install bubblewrap
+  when: ansible_os_family == 'RedHat'
+  yum:
+    name:
+      - bubblewrap
```
jamielennox: does this even work? with_items or name: bubblewrap right?

jamielennox: can just use package: for this because same name across dists.

roles/logrotate/tasks/main.yml
```diff
@@ -1,9 +1,15 @@
...
 - name: Install base package
+  when: ansible_os_family == 'Debian'
   apt:
     name: logrotate
     state: installed
 
+- name: Install base package
+  when: ansible_os_family == 'RedHat'
+  yum:
+    name: logrotate
```
jamielennox, line 11: package:

roles/filebeat/tasks/main.yml
```diff
@@ -1,9 +1,15 @@
...
 - name: Install Filebeat.
+  when: ansible_os_family == 'Debian'
   apt:
     name: filebeat
     state: present
 
+- name: Install Fielbeat.
+  when: ansible_os_family == 'RedHat'
+  yum:
+    name: filebeat
```
jamielennox, line 11: package:

roles/integration-handler/handlers/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: Reload integration-handler uwsgi
   file:
-    path: /etc/uwsgi-emperor/vassals/integration-handler.ini
+    path: "{{ '/etc/uwsgi-emperor/vassals' if ansible_os_family == 'Debian' else '/etc/uwsgi.d' }}/integration-handler.ini"
```
jamielennox, line 4: it seems weird to me that this isn't handled by the role dependency

roles/nodepool/tasks/main.yml
```diff
@@ -93,8 +106,22 @@
   synchronize:
     src: files/etc/nodepool/
     dest: /etc/nodepool/
-    rsync_opts:
-      - "--chown=nodepool:nodepool"
+# The option below is amazing, but requires rsync 3.1+
+#    rsync_opts:
+#      - "--chown=nodepool:nodepool"
```
jamielennox, line 111: we should look at pulling these out into their own git repo

roles/nodepool/tasks/main.yml
```diff
@@ -93,8 +106,22 @@
...
+#
+# So we need to manually do the same thing after the fact
+
+- name: Get list of files in /etc/nodepool
+  command: find /etc/nodepool -type f \! -name nodepool.yaml -a \! -name secure.conf
+  register: files_in_etc_nodepool
```
jamielennox, line 117: why aren't nodepool.yaml and secure.conf owned by nodepool anyway?

jamielennox: Feels like there is a better way we could organize all this stuff, maybe put nodepool templates in opt?

roles/zookeeper/files/usr/lib/systemd/system/zookeeper.service
```diff
@@ -5,9 +5,11 @@ After=syslog.target
...
 SyslogIdentifier=zookeeper
 TimeoutStartSec=10min
-Type=forking
-ExecStart=/usr/share/zookeeper/bin/zkServer.sh start
-ExecStop=/usr/share/zookeeper/bin/zkServer.sh stop
+User=zookeeper
+Group=zookeeper
+WorkingDirectory=/var/lib/zookeeper
+Environment="ZOOCFGDIR=/etc/zookeeper/conf"
+ExecStart=/usr/share/zookeeper/bin/zkServer.sh start-foreground
```
jamielennox, line 12: squash

roles/apache/tasks/main.yml
SpamapS: I'll rename to redhat.yaml to match the later patches (as you can tell this was an earlier patch. Also later I abandoned apache and used nginx, so I'm not 100% sure this even works.. I'm also not sure why we have apache still)

roles/apache/tasks/ssl_redhat.yml
SpamapS: I like that approach too. This is lazy, but it does work.

roles/common/tasks/main.yml
SpamapS: Oooops forgot the when! :) As far as using 'package:' ... at what point does it become unreadable and you have to dig in 3 different files to find the list of packages?

roles/sshd/tasks/main.yml
SpamapS: When we add a third distro, we should totally do that. ;)

roles/apache/tasks/apache_redhat.yml
SpamapS: Done

roles/mysql/tasks/debian.yml
SpamapS: I don't actually know how to do that. An example of how you'd like it done would be helpful, thanks.

roles/zookeeper/tasks/upstream_install.yml
SpamapS: It is because it is version specific. RPMs would require repos, and I don't have one that is reliable. You got one? I'll fix this happily!

roles/pip/vars/main.yml
SpamapS: It's because of the altinstall. We keep the system level pip executable.

roles/sshd/handlers/main.yml
SpamapS: Done

roles/python-app/vars/main.yml
SpamapS: Done

roles/nginx/tasks/redhat.yml
SpamapS: Yum works better if you pass all the packages in at once. Apt doesn't allow this, but yum does.

roles/python-app/vars/main.yml
SpamapS: It will only cause problems when you go to 3.6. A bridge to cross later?

roles/zuul-executor/tasks/main.yml
SpamapS: TIL about package:

roles/nodepool/tasks/main.yml
SpamapS: Agreed. This solution, btw, SUCKS. It changes things every time. I'd much rather use --chown. I might even be willing to chase down upgrade rsync for centos 7. But.. this was the quickest way I got to a working nodepool on a short schedule. Open to other suggestions too.

roles/zookeeper/files/usr/lib/systemd/system/zookeeper.service
SpamapS: Done

SpamapS:
FYI, I did some squashes so it's at least a few less patches. No less daunting as a review task though. Sorry for that.

